PR #3, opened 2025-05-07
Add versioned release workflow

dkhofer:


dkhofer:
I haven't been able to fully test this because the action that updates the cargo version can tell that I'm running it as part of a pull request and doesn't actually update the version.  The yml is valid but it seems likely I'll need to iterate, if I hit any errors with the git commands

Chris7:
looks good to me, hope it works on main.Feel free to just push as much as you want there to get it right.

Chris7 (approved):

PR #4, opened 2025-05-09
Handle gzip better

Chris7:
noodles-bgzf chokes when a normal gzipped fasta is passed. Makes a better handler for this to manage all 3 cases. We likely should factor this somewhere more common at som epoint.


dkhofer (approved):
LGTM, thanks!

PR #7, opened 2025-09-26
Add python bindings for some commands

dkhofer:


gen-core/src/lib.rs
```diff
@@ -92,7 +92,7 @@ impl<'de> serde::Deserialize<'de> for HashId {
 impl fmt::Display for HashId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         for byte in &self.0 {
-            write!(f, "{:02x}", byte)?;
+            write!(f, "{byte:02x}")?;
```
dkhofer, line 95: Just pacifying clippy with a lot of this

PR #9, opened 2025-09-26
Update dependencies

Chris7:
Resolves the slab 4.10 security advisory and just gets us up to date on a handful of things.


gen-core/Cargo.lock
Chris7, line 1: shouldn't have these apparently

dkhofer (approved):
LGTM, thanks!

PR #10, opened 2025-09-26
Pluralize operation and file_addition tables

dkhofer:


gen-core/src/lib.rs
```diff
@@ -92,7 +92,7 @@ impl<'de> serde::Deserialize<'de> for HashId {
 impl fmt::Display for HashId {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         for byte in &self.0 {
-            write!(f, "{:02x}", byte)?;
+            write!(f, "{byte:02x}")?;
```
dkhofer, line 95: Some clippy appeasement here

Chris7: yeah these are definitely better this way. Surprised it let me get by

Chris7 (approved):
lgtm

PR #11, opened 2025-09-27
Manifest for local transfers

Chris7:


src/main.rs
```diff
@@ -59,6 +59,7 @@ fn main() {
     // commands not requiring a db connection are handled here
     if let Some(Commands::Init {}) = &cli.command {
         get_or_create_gen_dir();
+        get_operation_connection(None).unwrap();
```
Chris7, line 62: init functionality should make the gen.db

dkhofer (approved):
LGTM.  Basically this shifts my (very rough) code to transfer at the file level to instead transfer at the changeset level, a welcome change.

PR #15, opened 2025-09-30
File addition checksum

Chris7:


gen-models/src/operations.rs
```diff
@@ -233,11 +239,28 @@ pub struct OperationInfo {
...
+    let mut hasher = Sha256::new();
+    io::copy(&mut reader, &mut hasher);
+    let result = hasher.finalize();
+    let mut hash_array = [0u8; 32];
+    hash_array.copy_from_slice(&result);
+    Ok(hash_array)
+}
+
 #[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
 pub struct FileAddition {
```
Chris7, line 259: we store a hash on both id and checksum. This is so empty files don't all have the same pointer, it's unique by (file_path, checksum)

gen-models/gen-models.capnp
```diff
@@ -130,9 +130,10 @@ struct Operation {
 }
 
 struct FileAddition {
-  id @0 :Int64;
+  id @0 :List(UInt8);
```
dkhofer, line 133: I've been thinking some of the other IDs we have should probably get converted from eg i64 to u64

dkhofer (approved):
LGTM, comments are nonblocking

PR #16, opened 2025-10-06
WIP: Viewer refactor

bobvh:
Update and reimport legacy code to fit in new structure Implemented HashId
Fixed Rust2024 match errors
There's still a bug when zooming in/out on partitions that are not the origin, hence excess logging in graph_controller.rs. I think it's the order of operations when you change the anchor partition that's to blame, but I wanted to get this on the repo before Monday.


bobvh:
(Q: Why the force-push twice? 
 A: I wanted to link Chris to a specific commit, but then noticed that the commit message had a copy-paste error in it that I wanted to rectify. Learned my lesson about changing history on a commit that happened before a merge commit.)

gen-annotations/Cargo.toml
```diff
@@ -2,13 +2,16 @@
 name = "gen-annotations"
 version = "0.1.0"
 edition = "2024"
+license = "Apache-2.0"
 
 [dependencies]
+anyhow = "1.0"
```
Chris7: why did these get added?

PR #18, opened 2025-10-07
Add in ability to delete part of a sequence via update fasta

dkhofer:


src/updates/fasta.rs
```diff
@@ -69,73 +69,113 @@ pub fn update_with_fasta(
...
-            preserve_edge: true,
-        };
-        BlockGroup::insert_change(conn, &path_change, &interval_tree).unwrap();
-        if index == 0 {
-            first_node = Some(node_id);
-        } else if first_node.is_some() {
-            first_node = None;
+        if sequence.is_empty() {
+            // We assume this is a deletion.
+            let node_id = HashId::convert_str("");
```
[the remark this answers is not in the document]
Chris7: Maybe use the same hash as above w/ sequence_hash being the blank?

src/updates/fasta.rs
```diff
@@ -69,73 +69,113 @@ pub fn update_with_fasta(
...
-            &edge_to_new_node,
-            &edge_from_new_node,
-        );
+        if node_id == HashId::convert_str("") {
+            let _ = path.new_path_with_deletion(conn, start_coordinate, end_coordinate);
+        } else {
+            let edge_to_new_node = Edge::query(
+                conn,
+                "select * from edges where target_node_id = ?1",
+                rusqlite::params!(SQLValue::from(node_id)),
```
Chris7, line 162: you can actually rewrite this just as `params![node_id]`, no need to import SQLValue.

gen-models/src/path.rs
```diff
@@ -672,6 +673,91 @@ impl Path {
...
+            .map(|edge| ((edge.source_node_id, edge.source_coordinate), edge))
+            .collect::<HashMap<(_, i64), &Edge>>();
+        let edges_by_target = edges
+            .iter()
+            .map(|edge| ((edge.target_node_id, edge.target_coordinate), edge))
+            .collect::<HashMap<(_, i64), &Edge>>();
+        let edge_before_deletion = edges_by_target
+            .get(&(block_with_start.node_id, block_with_start.sequence_start))
+            .unwrap();
+        let edge_after_deletion = edges_by_source
```
Chris7, line 723: what happens on the extremes where there are no edges after the deletion?

Chris7 (commented):
I think this will fail on the sequence extremes by the unwrap call. The other thing I'd like clarity on is the empty nodeid reuse, is that intentional (vs. using the path_id:start-end->"" syntax)

PR #75, opened 2026-03-06
Import genbank annotations

Chris7:
Support for importing annotations w/ genbank files

### DIFF
--- a/gen-annotations/src/translate/bed.rs
+++ b/gen-annotations/src/translate/bed.rs
@@ -39,9 +39,9 @@ where
     R: Read,
     W: Write,
 {
-    let mut record = bed::Record::default();
-    let mut bed_reader = bed::io::reader::Builder::<3>.build_from_reader(reader);
-    let mut bed_writer = bed::io::Writer::<3, _>::new(writer);
+    let mut record = bed::Record::<6>::default();
+    let mut bed_reader = bed::io::reader::Builder::<6>.build_from_reader(reader);
+    let mut bed_writer = bed::io::Writer::<6, _>::new(writer);
 
     let bgs = Sample::get_block_groups(conn, collection, sample);
     let sample_bgs: HashMap<String, &BlockGroup> = HashMap::from_iter(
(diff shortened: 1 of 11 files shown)


gen-annotations/src/translate/bed.rs
```diff
@@ -39,9 +39,9 @@ where
     R: Read,
     W: Write,
 {
-    let mut record = bed::Record::default();
```
Chris7, line 42: switch to read bed6 for strand

src/imports/genbank.rs
```diff
@@ -17,16 +24,430 @@ use gen_models::{
...
+                    "{}:{}:{}:{}",
+                    segment.node_id, segment.sequence_start, segment.sequence_end, segment.strand
+                )
+            })
+            .collect::<Vec<_>>()
+            .join("|");
+        let annotation_sequence_hash = HashId::convert_str(&format!(
+            "{annotation_sequence_key};mapped_segments={mapped_segment_key}"
+        ));
+        let accession_name = format!(
```
Chris7, line 425: this is the accession_name, not the annotation_name. Accession_names are unique by their name which contribute to the id. So this won't be how the annotation names appear. The more I think about it the more I think accession should not be a user facing feature and something that something else like annotations strictly wraps.

src/genbank.rs
```diff
@@ -118,6 +142,96 @@ impl GenBankLocus {
...
+        .find_map(|(qualifier_key, value)| {
+            let qualifier_name: &str = qualifier_key.as_ref();
+            qualifier_name
+                .eq_ignore_ascii_case(key)
+                .then(|| value.as_deref().map(str::trim).map(str::to_string))
+                .flatten()
+        })
+}
+
+fn merge_annotation_segments(
```
Chris7, line 158: this one looks like the other one but different types

gen-annotations/src/translate/bed.rs
dkhofer: :+1:

src/imports/genbank.rs
dkhofer: :+1:

src/genbank.rs
dkhofer: Uh, which other one?

dkhofer: The one in the import code?

dkhofer (approved):
This all seems reasonable.  LGTM

PR #80, opened 2026-03-20
Refactor update with library so it can be called from python with no …

dkhofer:
…intermediate files

Example usages:

```
import gen
gen.init()
db_context = gen.DbContext()

from gen import PySequencePart
part1 = PySequencePart("foo", "AAAA", 4)
part2 = PySequencePart("bar", "CCCC", 4)
part3 = PySequencePart("baz", "GGGG", 4)
part4 = PySequencePart("quux", "TTTT", 4)

import_library(dbc, "design", [[part1, part2], [part3, part4]], None, None)
```

```
import gen
gen.init()
db_context = gen.DbContext()

gen.import_fasta(dbc, "/Users/hofer/github/genhub/gen/fixtures/aaaaaaaa.fa", None, None, False)
gen.update_with_library(dbc, None, None, "design", "m123", 2, 6, [[part1, part2], [part3, part4]])
```


gen-python/src/python_api/sequence_part.rs
```diff
@@ -0,0 +1,22 @@
...
+pub struct PySequencePart {
+    pub name: String,
+    pub sequence: String,
+    pub sequence_length: i64,
+}
+
+#[pymethods]
+impl PySequencePart {
+    #[new]
+    fn new(name: String, sequence: String, sequence_length: i64) -> Self {
```
Chris7: can we have the length be determined from the sequence if omitted on the python api side?

gen-python/src/imports.rs
```diff
@@ -178,9 +188,78 @@ pub fn import_library(
...
+    let conn = context.graph().conn();
+
+    if let Err(err) = r#gen::track_database(conn, operation_conn) {
+        panic!("Error tracking database: {err}");
+    }
+
+    conn.execute("BEGIN TRANSACTION", []).unwrap();
+    operation_conn.execute("BEGIN TRANSACTION", []).unwrap();
+
+    let name = name.unwrap_or_else(|| r#gen::commands::get_default_collection(operation_conn));
```
Chris7: we should move the get_default_collection to models. It gives me a bad feeling having command imports in here.

gen-python/Cargo.toml
```diff
@@ -9,7 +9,7 @@ license = "Apache-2.0"
 publish = false
 
 [lib]
-name = "gen_python"
```
Chris7, line 12: why was this renamed?

src/graphs/combinatorial_library.rs
```diff
@@ -126,6 +126,14 @@ pub fn create_library(
         parts_set.extend(parts);
     }
 
+    let mut cleaned_parts_list = vec![];
```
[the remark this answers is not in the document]
Chris7: does order matter? (wrt using a hashset)

Chris7 (approved):
looks reasonable. Few questions that may be minor. I think having the define the sequence length always on the python side would be a pain because it is something we can automate and do on the rust side w/ the sequence builder.

gen-python/Cargo.toml
dkhofer: Reverted after asking Bob

src/graphs/combinatorial_library.rs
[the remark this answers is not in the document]
dkhofer: I think so, that's why I'm using the retain method, I think inserting into a HashSet returns true if the value is freshly inserted, false if it's already present.

I'm assuming users will expect to see their parts in the order they provided them

gen-python/src/python_api/sequence_part.rs
dkhofer: Good idea, done

gen-python/src/imports.rs
dkhofer: I found all the places the command is being called and replaced it with a Defaults method call

PR #81, opened 2026-03-24
Sample lineage

Chris7:
Sample parent lineage. Implemented as a general trait since we have  few graph-in-sql structures.

* Duplicate was removed and changed to 2 functions -- one for merging the graphs, another for merging in paths/accessions. It was done to handle the case of multiple parents.
* Other related changes to having multiple parent samples


gen-models/src/block_group.rs
```diff
@@ -287,37 +272,87 @@ impl BlockGroup {
...
+        );
 
-        if let Some(parent_block_group) = block_groups.first() {
+        if let Some(parent_block_group) = parent_block_groups.first() {
             let new_block_group =
                 BlockGroup::create(conn, collection_name, sample_name, group_name);
+            let mut merged_graph = GenGraph::new();
 
-            // clone parent blocks/edges/path
-            parent_block_group.duplicate(conn, &new_block_group.id);
```
Chris7, line 306: duplicate has been folded into merge_from since it was the same behavior and now we're combining multiple parents it's not really duplicate anymore anyways.

gen-models/src/block_group.rs
```diff
@@ -203,39 +208,25 @@ impl BlockGroup {
         }
     }
 
-    pub fn duplicate(&self, conn: &GraphConnection, target_block_group_id: &HashId) {
```
Chris7, line 206: duplicate replaced with a merge function below to handle multiple parents

gen-models/src/block_group.rs
```diff
@@ -203,39 +208,25 @@ impl BlockGroup {
...
+    fn copy_paths_and_accessions_into(
```
Chris7, line 273: the path/accession part of duplicate is here, so with each parent we can copy its paths/accessions into the child

gen-models/src/block_group.rs
```diff
@@ -135,10 +135,15 @@ impl<'a> PathCache<'a> {
     }
 
     pub fn get_intervaltree<'b>(
```
Chris7, line 183: this function only worked with a preloaded cache, so cache was made mutable so it can insert entries

dkhofer: :+1:

gen-models/src/block_group.rs
dkhofer: :+1:

gen-models/src/block_group.rs
dkhofer: :+1:

gen-models/src/block_group.rs
dkhofer: :+1:

gen-models/src/block_group.rs
```diff
@@ -1229,12 +1270,320 @@ mod tests {
...
+
+        let merged_graph = BlockGroup::get_graph(conn, &merged_bg_id);
+        let merged_node_ids = merged_graph
+            .nodes()
+            .filter(|node| !is_terminal(node.node_id))
+            .map(|node| node.node_id)
+            .collect::<HashSet<_>>();
+
+        assert_eq!(merged_node_ids, HashSet::from([node_a, node_b]));
+    }
```
dkhofer, line 1544: Might also be good to add the standard `get_all_sequences` check here

gen-models/src/lineage.rs
```diff
@@ -0,0 +1,558 @@
...
+        .collect()
+}
+
+impl LineageId for String {
+    fn decode_hex_token(token: &str) -> Self {
+        String::from_utf8(decode_hex_bytes(token)).expect("lineage token should decode to utf-8")
+    }
+}
+
+impl LineageId for HashId {
```
dkhofer, line 37: Can you explain more about why we're storing the lineage ID in hex, and then converting to a hash ID?  Maybe add a comment?

gen-models/src/lineage.rs
```diff
@@ -0,0 +1,558 @@
...
+                    <$ty>::from_str(&value).expect("lineage token should decode to a number")
+                }
+            }
+        )*
+    };
+}
+
+impl_numeric_lineage_id!(i32, i64, u32, u64, usize);
+
+pub trait SqlLineage: Query<Model = Self> + Sized {
```
dkhofer, line 60: This is a great thing to have implemented as a trait, thanks!

dkhofer (approved):
Looks reasonable, LGTM, just have a few nonblocking questions and comments.

gen-models/src/block_group.rs
Chris7: That actually raised a good bug. When merging graphs, we need to offset the chromosome_index of the incoming graph if there are conflicts, otherwise those edges get pruned! Thanks for the suggestion.

PR #82, opened 2026-03-25
Remove block id

Chris7:
With the node/etc. changes it seems we no longer need block ids. Tests pass and it looks good.

### DIFF
--- a/gen-annotations/src/translate/bed.rs
+++ b/gen-annotations/src/translate/bed.rs
@@ -80,6 +80,8 @@ where
 mod tests {
     use std::{fs::File, path::PathBuf};
 
+    use gen_models::sample::Sample;
+
     use super::translate_bed;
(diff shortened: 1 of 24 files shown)


gen-annotations/src/translate/bed.rs
```diff
@@ -80,6 +80,8 @@ where
 mod tests {
     use std::{fs::File, path::PathBuf};
 
+    use gen_models::sample::Sample;
```
dkhofer: Why is this import getting added when nothing else is changing in this file (or Sample)?  Same question for gff.rs below

dkhofer (approved):
LGTM.  My suspicion is that before we had nodes, I needed a way to map from an intervaltree to a block, and I used the block ID as an intermediary value or something.  The (node ID, start, end) tuple works better though

PR #83, opened 2026-03-27
Fix path creation bug when updating with sequence at start or end of …

dkhofer:
…the graph

### DIFF
--- a/gen-models/src/path.rs
+++ b/gen-models/src/path.rs
@@ -3,7 +3,7 @@ use std::collections::{HashMap, HashSet};
 
 use gen_core::{
     HashId, NodeIntervalBlock, PATH_END_NODE_ID, PATH_START_NODE_ID, PathBlock, Strand,
-    calculate_hash,
+    calculate_hash, is_end_node, is_start_node,
     range::{Range, RangeMapping},
     traits::Capnp,
 };
@@ -293,10 +293,10 @@ impl Path {
 
         let mut sequence_node_ids = HashSet::new();
         for edge in &edges {
-            if edge.source_node_id != PATH_START_NODE_ID {
+            if !is_start_node(edge.source_node_id) {
                 sequence_node_ids.insert(edge.source_node_id);
             }
-            if edge.target_node_id != PATH_END_NODE_ID {
+            if !is_end_node(edge.target_node_id) {
                 sequence_node_ids.insert(edge.target_node_id);
             }
         }
--- a/gen-models/src/path.rs
+++ b/gen-models/src/path.rs
@@ -618,29 +618,46 @@ impl Path {
             .iter()
             .map(|edge| ((edge.target_node_id, edge.target_coordinate), edge))
             .collect::<HashMap<(_, i64), &Edge>>();
-        let edge_before_new_node = edges_by_target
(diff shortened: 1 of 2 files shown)


dkhofer:
The cargo update in gen-python is to resolve this security warning: https://github.com/genhub-bio/gen/security/dependabot/33

gen-models/src/path.rs
```diff
@@ -618,29 +618,46 @@ impl Path {
...
             .map(|edge| ((edge.target_node_id, edge.target_coordinate), edge))
             .collect::<HashMap<(_, i64), &Edge>>();
-        let edge_before_new_node = edges_by_target
-            .get(&(block_with_start.node_id, block_with_start.sequence_start))
-            .unwrap();
-        let edge_after_new_node = edges_by_source
-            .get(&(block_with_end.node_id, block_with_end.sequence_end))
-            .unwrap();
+
+        let edge_before_new_node = if edge_to_new_node.source_node_id == PATH_START_NODE_ID {
```
Chris7, line 622: can we use is_start_node/is_end_node?

Chris7: .

gen-models/src/path.rs
```diff
@@ -618,29 +618,46 @@ impl Path {
             .iter()
             .map(|edge| ((edge.target_node_id, edge.target_coordinate), edge))
             .collect::<HashMap<(_, i64), &Edge>>();
-        let edge_before_new_node = edges_by_target
```
Chris7, line 621: i think the edges_by_soruce/edges_by_target might be wrong here as well. We don't enforce uniqueness on the source_node/coordinate alone. So we can have multiple edges coming from the same source + going to the same target, but the way the hashmap works the last one wins. It should be a HashMap<coords, Vec<&Edge>>

gen-models/src/path.rs
```diff
@@ -618,29 +618,46 @@ impl Path {
...
-            .get(&(block_with_start.node_id, block_with_start.sequence_start))
-            .unwrap();
-        let edge_after_new_node = edges_by_source
-            .get(&(block_with_end.node_id, block_with_end.sequence_end))
-            .unwrap();
+
+        let edge_before_new_node = if edge_to_new_node.source_node_id == PATH_START_NODE_ID {
+            None
+        } else {
+            let edge = edges_by_target
```
Chris7, line 625: i think this should be edges

Chris7 (approved):
approach makes sense, but it hink we have another bug

gen-models/src/path.rs
dkhofer: I'm confused.  When would a path have duplicate edges?  That would imply a cycle somewhere.  There is some validation code here: https://github.com/genhub-bio/gen/blob/main/gen-models/src/path.rs#L110  The target of one edge must be the source of the next, so at least there can't be two consecutive duplicate edges.

gen-models/src/path.rs
dkhofer: Yep, changed

gen-models/src/path.rs
Chris7: you're right. i was thinking of the general model and not path specifics.

PR #88, opened 2026-04-02
Fix annotation <-> reference name translation

### LINKED ISSUE #85: Translating block group names
Working with yeast genomes from SGD and the 1011 yeast genomes project is a pain because, even though they're based on a known common reference:
- the reference strain from the SGD release has these fasta headers `ref|NC_001133| [org=Saccharomyces cerevisiae] [strain=S288C] [moltype=genomic] [chromosome=I]` , what ends up in gen is `ref|NC_001133|`.
- the GFF annotations from SGD use `chrI` as contig name
- the VCF files from 1011 yeast genome use `chromosome1`

We discussed starting to build up a ruleset so we can do this automatically as a fallback (with a message to the user and a way to disable it). We could also support the other side of the spectrum by having an option for a translation table that users would provide themselves?

dkhofer:
Resolves https://github.com/genhub-bio/gen/issues/85


gen-models/migrations/core/01-initial/up.sql
```diff
@@ -126,3 +132,16 @@ INSERT INTO gen_metadata (db_uuid) values (lower(
 INSERT INTO sequences (hash, sequence_type, sequence, name, file_path, "length") values (X'84d6adbd5395281933fe41e877d3a7f02a3b1990a65be1901b2c91fc685e083b', "OTHER", "start-node-yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy", "", "", 64), (X'1c7dfc64977b0838af0762d7333dcb64c175b15e65a70099ec38f46bf1a15ea3', "OTHER", "end-node-zzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzzz", "", "", 64);
 INSERT INTO nodes (id, sequence_hash) values (X'84d6adbd5395281933fe41e877d3a7f02a3b1990a65be1901b2c91fc685e083b', X'84d6adbd5395281933fe41e877d3a7f02a3b1990a65be1901b2c91fc685e083b');
 INSERT INTO nodes (id, sequence_hash) values (X'1c7dfc64977b0838af0762d7333dcb64c175b15e65a70099ec38f46bf1a15ea3', X'1c7dfc64977b0838af0762d7333dcb64c175b15e65a70099ec38f46bf1a15ea3');
+
+INSERT INTO reference_aliases (reference_name, refseq_accession_id, genbank_accession_id) values ('E. coli K-12 MG1655', 'NC_000913.3', 'U00096.3');
```
Chris7: is "reference name" our attempt to do a normalized name?

Chris7: How should we support names that map to something like "chr1" and "1", "I"? A column called misc that is a json array of other names to load?

dkhofer: "Reference name" is there because I wanted a human-readable field so it's clearer what organism the row represents.  I'm just using whatever ncbi is providing.

For this PR I'm focused on solving the fasta name <-> GFF name mapping, so that's why I've got just two fields beyond reference name.  If we want the UCSC reference name or some other field, or a misc field, I'm open to that, but not planning to do that here unless there's evidence of common GFF usage of other reference names for things like e coli, yeast, mouse, human.

dkhofer: Actually, looks like while my issue with e coli is solved by associating refseq and genbank names, Bob's issue requires the UCSC name so I'll add that in, and maybe a misc field too

Chris7: Here's some common stuff we'll hit:
* Import a genome fasta from refseq/ncbi -- that will have [NC_000001.11](https://www.ncbi.nlm.nih.gov/nuccore/NC_000001.11/) for human's chr1
* Use the GFF from GenCode -- https://ftp.ebi.ac.uk/pub/databases/gencode/Gencode_human/release_49/gencode.v49.annotation.gff3.gz. That will have Chr1 like:
```
##gff-version 3
#description: evidence-based annotation of the human genome (GRCh38), version 49 (Ensembl 115)
#provider: GENCODE
#contact: gencode-help@ebi.ac.uk
#format: gff3
#date: 2025-07-08
##sequence-region chr1 1 248956422
chr1    HAVANA  gene    11121   24894   .       +       .       ID=ENSG00000290825.2;gene_id=ENSG00000290825.2;gene_type=lncRNA;gene_name=DDX11L16;level=2;tag=overlaps_pseudogene
chr1    HAVANA  transcript      11121   14413   .       +       .       ID=ENST00000832824.1;Parent=ENSG00000290825.2;gene_id=ENSG00000290825.2;transcript_id=ENST00000832824.1;gene_type=lncRNA;gene_name=DDX11L16;transcript_type=lncRNA;transcript_name=DDX11L16-260;level=2;tag=TAGENE
chr1    HAVANA  exon    11121   11211   .       +       .       ID=exon:ENST00000832824.1:1;Parent=ENST00000832824.1;gene_id=ENSG00000290825.2;transcript_id=ENST00000832824.1;gene_type=lncRNA;gene_name=DDX11L16;transcript_type=lncRNA;transcript_name=DDX11L16-260;exon_number=1;exon_id=ENSE00004248723.1;level=2;tag=TAGENE
chr1    HAVANA  exon    12010   12227   .       +       .       ID=exon:ENST00000832824.1:2;Parent=ENST00000832824.1;gene_id=ENSG00000290825.2;transcript_id=ENST00000832824.1;gene_type=lncRNA;gene_name=DDX11L16;transcript_type=lncRNA;transcript_name=DDX11L16-260;exon_number=2;exon_id=ENSE00004248735.1;level=2;tag=TAGENE
chr1    HAVANA  exon    12613   12721   .       +       .       ID=exon:ENST00000832824.1:3;Parent=ENST00000832824.1;gene_id=ENSG00000290825.2;transcript_id=ENST00000832824.1;gene_type=lncRNA;gene_name=DDX11L16;transcript_type=lncRNA;tr
```

Chris7: Vcfs will be similar w/ Chr1.

dkhofer:
Seems to work.  Commands using Bob's yeast files, unmodified:

```
gen init
gen import fasta --sample reference S288C_reference_sequence_R64-1-1_20110203.fsa
gen add-annotation-file saccharomyces_cerevisiae_R64-1-1_20110208.gff
gen update vcf --parent-sample reference BRQ_trunk.vcf
```

gen-models/migrations/core/01-initial/up.sql
dkhofer: OK, yeah, this is more of an attempt to do normalized name now.  For fields I've now got refseq ID, genbank ID, ensembl ID, UCSC ID, "misc" (not used yet, can be null), and "chromosome" which is a nullable int intended to be the chromosome number.  So for yeast, everything seems to be roman numerals but the vcf can still be regular numbers (eg chrX versus chr10), so for yeast I'm manually populating the chromosome field to be for instance 10 if the chromosome is chrX.  More details on how I'm using "10" below, but first the high level explanation.

Basically for each reference name, I'm building a list of strings that could be aliases, plus the reference name.  For each block group name, if it's in the alias list, then I add each alias as a key in a hashmap pointing back to the block group name as a value.  Then when parsing GFF or VCF or whatever, if the seq/ref name is present as a key in the alias map, we use the value as the block group to apply the annotation or variant to.

To build the list of aliases, I'm defining a compute_aliases method that uses the reference alias fields and constructs some additional strings in memory that could plausibly also be a name for the same reference contig, and returns it all in a big hashset.  (So if the chromosome field is 10, we also add "chr10", "Chr10", "chromosome10", etc.)  That contains the messiness to one method.  The resulting hashset is what I use to populate the alias -> reference hashmap.

gen-models/src/reference_alias.rs
```diff
@@ -0,0 +1,216 @@
...
+    pub fn get_references_by_alias(
+        conn: &GraphConnection,
+        references: Vec<String>,
+    ) -> Result<HashMap<String, String>> {
+        let mut references_by_alias = HashMap::new();
+        let reference_aliases = ReferenceAlias::load_all(conn)?;
+        for reference_alias in reference_aliases {
+            let aliases = ReferenceAlias::compute_aliases(reference_alias);
+            for reference in &references {
+                if aliases.contains(reference) {
```
dkhofer, line 127: This could be problematic if there is somehow overlap between two sets of reference aliases

gen-annotations/src/translate/bed.rs
```diff
@@ -35,10 +38,19 @@ where
             .map(|bg| (bg.name.clone(), bg))
             .collect::<Vec<(String, &BlockGroup)>>(),
     );
+
+    // Load all reference aliases, to accommodate alternate reference names in the GFF file
+    let references = sample_bgs.keys().cloned().collect::<Vec<String>>();
+    let references_by_alias = ReferenceAlias::get_references_by_alias(conn, references)?;
```
Chris7, line 55: if we scoped referencealias by blockgroup, we could figure out chr1 is actually NC_00000.1 as well. Then in the lineage code we can copy down aliases to keep it working as usual.

gen-annotations/src/translate/bed.rs
```diff
@@ -1,12 +1,15 @@
 use std::{
     cmp::{max, min},
     collections::HashMap,
-    io::{Error, Read, Write},
+    io::{Read, Write},
 };
 
+use anyhow::Result;
```
Chris7: Anyhow is for application code. For errors, we should make a new TranslateBedError in errors.rs using thiserror and specify errors that way.

gen-models/migrations/core/01-initial/up.sql
```diff
@@ -116,6 +116,16 @@ CREATE TABLE block_group_edges (
 ) STRICT;
 CREATE UNIQUE INDEX block_group_edges_uidx ON block_group_edges(block_group_id, edge_id, chromosome_index, phased);
 
+CREATE TABLE reference_aliases (
```
Chris7: I've been adding new migrations these days. Can you make this core/0x-reference/up.sql?

Chris7: need a down.sql too

gen-models/src/reference_alias.rs
```diff
@@ -0,0 +1,216 @@
...
+        chromosome: Option<i64>,
+    ) -> Result<()> {
+        conn.execute(
+            "INSERT INTO reference_aliases (reference_name, refseq_accession_id, genbank_accession_id, ucsc_id, ensembl_id, custom_id, chromosome) VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7)",
+            rusqlite::params![reference_name, refseq_accession_id, genbank_accession_id, ucsc_id, ensembl_id, custom_id, chromosome],
+        )?;
+        Ok(())
+    }
+
+    pub fn load_all(conn: &GraphConnection) -> Result<Vec<ReferenceAlias>> {
```
Chris7: We get this through `all` in the Query trait

gen-models/src/reference_alias.rs
```diff
@@ -0,0 +1,216 @@
...
+            let genbank_without_version = reference_alias
+                .genbank_accession_id
+                .split('.')
+                .next()
+                .unwrap();
+            aliases.insert(genbank_without_version.to_string());
+        }
+        aliases.insert(reference_alias.ucsc_id);
+        aliases.insert(reference_alias.ensembl_id.clone());
+        aliases.insert(format!("chr{}", reference_alias.ensembl_id));
```
Chris7: we should make downstream stuff case insensitive vs. this

src/commands/mod.rs
```diff
@@ -361,6 +361,29 @@ pub enum Commands {
         #[arg(long)]
         new_region: String,
     },
+    AddReferenceAliases {
+        /// The name of the reference to set up aliases for
+        #[arg(long)]
+        reference_name: String,
+        /// The refseq accession ID
+        #[arg(long)]
+        refseq_accession_id: String,
```
Chris7: we should have optional for quite a few of these. It's not guaranteed that an organism is in genbank/refseq/ensembl/ucsc

Chris7: do we actually care about the soruce as well? Could we do this as (reference_name, alias) and just ensure that's a unique pair? With my prior suggestion it would be (bg_id, reference_name, alias)

Chris7 (commented):
I have a few thoughts

src/commands/mod.rs
Chris7: i guess an advantage of this is we know ucsc id X == ensembl id Y without defining all the cross pairs

gen-annotations/src/translate/bed.rs
dkhofer: That is a decent idea, and I thought about it for a while, but I'm having trouble figuring out how to make it work, because I'm not sure what assumptions we can make about the schema and data.

I am a bit confused about what you mean by scoping referencealias by blockgroup.  I think the referencealias rows should exist independently of block groups.  I think the best solution would be a join table between block group and reference alias.  But then the question becomes which referencealias fields we put in the join table.  I think all the relevant ones could be nullable, except reference name, which may be duplicated (for organisms with multiple chromosomes).  I guess we could force reference name to end in something like "chr1" or something.  Or we could add an incrementing ID field to referencealias.

To me, the benefit of adding a join table would be to log which reference we've associated a block group with, and slightly faster lookup time for the reference aliases of a block group, although I'm not sure speed will be a huge issue here.  I'm not convinced those are important enough for us to nail down that part of the schema right now.

gen-annotations/src/translate/bed.rs
dkhofer: Oops, fixed

gen-models/migrations/core/01-initial/up.sql
dkhofer: Yep, added

dkhofer: Added

gen-models/src/reference_alias.rs
dkhofer: Oops, did that

gen-models/src/reference_alias.rs
dkhofer: I'd prefer to just keep it simple for now and not do downstream processing, is that all right?

src/commands/mod.rs
dkhofer: Yep, changed.  Also added a uniqueness constraint on the refseq ID.  Nothing else seems to be especially unique otherwise, unfortunately

dkhofer: Yeah, I like having it all in one row

dkhofer:
This is ready for another look

gen-models/src/reference_alias.rs
```diff
@@ -0,0 +1,211 @@
...
+            aliases.insert(format!("chromosome{}", ensembl_id));
+            aliases.insert(format!("Chromosome{}", ensembl_id));
+        }
+        if let Some(custom_id) = reference_alias.custom_id {
+            aliases.insert(custom_id.clone());
+            aliases.insert(format!("chr{}", custom_id));
+            aliases.insert(format!("Chr{}", custom_id));
+            aliases.insert(format!("chrom{}", custom_id));
+            aliases.insert(format!("Chrom{}", custom_id));
+            aliases.insert(format!("chr{}", custom_id));
```
Chris7: should be `chromosome`. Maybe a helper method to do all these?

Chris7 (approved):
One comment. I think at some point we should use something like caseless to store these strings so we can do case-insensitive comparisons w/o a speed hit.

PR #89, opened 2026-04-06
Path parent

Chris7:
Keeps a pointer to path parentage for use when cloning paths from parents. A new field, is_default, is present as well so users can define which parent path to default to for coordinates. This would be used in the cases when a multi-parent sample is created where the parents have different reference genomes for a given path (such as 2 strains of yeast being crossed and each strain may have a different reference chr1/etc.)

### DIFF
--- a/gen-models/src/block_group.rs
+++ b/gen-models/src/block_group.rs
@@ -1396,6 +1396,11 @@ mod tests {
             .collect::<HashSet<_>>();
 
         assert_eq!(merged_node_ids, HashSet::from([node_a, node_b]));
+        let all_sequences = BlockGroup::get_all_sequences(conn, &merged_bg_id, false);
+        assert_eq!(
+            all_sequences,
+            HashSet::from_iter(vec!["AAAA".to_string(), "CCCC".to_string()])
+        );
     }
 
     #[test]
--- a/gen-models/src/block_group.rs
+++ b/gen-models/src/block_group.rs
@@ -120,12 +120,8 @@ impl<'a> PathCache<'a> {
             path.clone()
         } else {
             let conn = path_cache.conn;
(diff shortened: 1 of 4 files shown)


gen-models/migrations/core/04-path-parent-lineage/down.sql
```diff
@@ -0,0 +1,51 @@
...
+  created_on INTEGER NOT NULL,
+  FOREIGN KEY(block_group_id) REFERENCES block_groups(id)
+) STRICT;
+
+INSERT INTO paths_rollback (id, block_group_id, name, created_on)
+SELECT id, block_group_id, name, created_on
+FROM paths
+WHERE is_default = 1;
+
+DELETE FROM accession_paths
```
Chris7: these deletions remove paths that will not be in the paths_rollback (any non-default paths)

gen-models/src/block_group.rs
```diff
@@ -215,17 +211,38 @@ impl BlockGroup {
...
+            .map(|path| path.name)
+            .collect::<HashSet<_>>();
         let mut path_map = HashMap::new();
         for path in &existing_paths {
             let edge_ids = PathEdge::edges_for_path(conn, &path.id)
                 .into_iter()
                 .map(|edge| edge.id)
                 .collect::<Vec<_>>();
-            let new_path = Path::create(conn, &path.name, target_block_group_id, &edge_ids);
+            let is_default = if default_path_names.contains(&path.name) {
```
Chris7: We're copying paths from the parent into the new block group. So here, we're saying:
* Does a default already exist in the target for this name?
* If no, is the current path a default, or do we have no parent path (in which case set this new path to default)

PR #90, opened 2026-04-08
Jupyter widget

bobvh:
Display the graph viewer in Jupyter notebooks and the preparations leading up to that. This branch doesn't use WASM, but if we ever want to, a WASM viewer isn't that far off. What it does do, is serialize a frame in Rust, pass it through Python/anywidget to a thin javascript layer that prints characters to a canvas and forwards keyboard/mouse events.

- Cleans up gen-tui to remove dependencies that were unnecessary coupling to GenGraph so that it can more easily be reused for other graphs.
- Gates Ratatui features to enable builds without terminal (crossterm) support.
- Mouse support for the CLI: drag to pan, scroll to zoom, click buttons in sidebar to switch back and forth between different block groups
- The Rust compiler doesn't allow Python objects to borrow from Rust since it wouldn't be able to guarantee lifetimes. The way the generic viewer was setup up before made it generic to borrowed or owned input data, here we force the GenGraph viewer implementation to use the owned version.
- blockgroup.plot() or display(blockgroup) to show a live, interactive graph browser in your Jupyter notebook. Controlled by mouse or Python methods. It uses anywidget and should probably work in similar notebook environments as well.


gen-python/python/gen/static/jupyter_widget.js
```diff
@@ -0,0 +1,205 @@
...
+const FONT = `${FONT_SIZE}px ${FONT_FAMILY}`;
+
+function fontForCell(cell) {
+  if (cell.bold && cell.italic) return `bold italic ${FONT_SIZE}px ${FONT_FAMILY}`;
+  if (cell.bold) return `bold ${FONT_SIZE}px ${FONT_FAMILY}`;
+  if (cell.italic) return `italic ${FONT_SIZE}px ${FONT_FAMILY}`;
+  return FONT;
+}
+
+function render({ model, el }) {
```
dkhofer, line 25: I'm guessing you're doing lowest common denominator javascript because it's not clear what else might be available in jupiter?  Are you sure there isn't a react/typescript like framework that comes with jupiter?  If not, I'm fine with what you've got here, it's fairly straightforward and I suspect a lot of the code would look similar even if put into framework methods

gen-tui/Cargo.toml
```diff
@@ -8,33 +8,24 @@ homepage = "https://genhub.bio"
 license = "Apache-2.0"
 
 [dependencies]
-gen-core = { path = "../gen-core", version = "0.1.30" }
-gen-models = { path = "../gen-models", version = "0.1.30" }
-gen-graph = { path = "../gen-graph", version = "0.1.30" }
 gen-sugiyama = { path = "../gen-sugiyama", version = "0.1.30" }
```
dkhofer, line 11: Just curious, is there also a plan to move the gen-sugiyama code to under gen-tui?  Seems like it and gen-python are the only subcrates that depend on it, and I suspect the gen-python dependency is because of gen-tui

src/views/block_group.rs
```diff
@@ -262,26 +262,325 @@ pub fn view_block_group(
...
     let mut last_selected_block_group_id = block_group_id;
     // Track if we're loading a new block group
     let mut is_loading = false;
     let mut last_refresh = Instant::now();
+    let mut should_quit = false;
     loop {
-        // Refresh explorer data and force reload on change
+        // Drain ALL pending input events before doing any work
+        while crossterm::event::poll(Duration::from_millis(0))? {
+            match event::read()? {
```
dkhofer, line 291: It might be good to refactor each of the blocks of code that handle event types into separate methods.  If we add more in the future, the amount of raw code inside the while loop could blow up, making it harder to understand the event handling at a high level

dkhofer (approved):
All seems reasonable.  Comments are nonblocking.  LGTM

dkhofer:
BTW just in case you weren't aware of the tradeoffs, the reason to pass references around instead of creating new objects and passing them (by value), is mainly for speed.  A reference is basically just a memory address, so sending it to a different method is cheaper than copying an entire object and passing that in.  But I'm pretty sure we don't need that level of optimization for UI-level stuff, given how fast rust is.  I guess we'll find out, but I'm not super worried

gen-tui/Cargo.toml
bobvh: I didn't realize that you could nest them multiple levels deep. That would definitely be the move, also because eventually I wanted to publish what's now called gen-tui to the Ratatui userbase as a generic widget.

bobvh: Another option was to check in with the status of the original sugiyama crate and see if we can work around the limitations now, even with a performance hit.

gen-python/python/gen/static/jupyter_widget.js
bobvh: That wasn't the reason. The browserside JS is 100% LLM product, and initially I had prompted for typescript to mirror what you and Chris do on the website. It worked fine, but my build process suddenly became a whole lot more complicated than I felt was justified for the minimal shell this is supposed to be (bringing in npm as a dependency etc).

bobvh:
Thanks! Turns out that I didn't actually need the graph after initializing the widget controller in the main viewer, so I'm moving it in now instead of moving its clone (GenGraph is not Copy). That also allowed me to cut out the Clone requirement altogether. So it's still not passing a reference, but it should be much cheaper than the clone. Is a move that much more expensive than passing a reference?

src/views/block_group.rs
bobvh: I agree that the event handling really needs to be dealt with in a more structured way, but we should change it for all the TUI components at once then (not just viewer)

gen-tui/Cargo.toml
dkhofer: I'm not sure if you can nest subcrates, but if the TUI is the only thing that uses sugiyama, you could just put the code into a subdirectory of gen-tui

PR #91, opened 2026-04-14
Blockgroup lineage

Chris7:
This is an add-on to #81 in response to the "we should use get_all_sequences" bit on a test. It adds a lineage to block groups so when a mating operation happens, there is a way to trace which block group came from which sample. I did it this way instead of the previous attempt to have a path lineage because of how w currently manage ploidyness w/ chromosome_index caused a lot of problems when we just mashed the block groups together. 

Some changes to note:
* When updating with a fasta/sequence/vcf/etc., we apply changes to all matching blockgroups. There's no disambiguation currently to just select the lineage of a blockgroup like parent:chr1 currently.
* get_or_create_sample_blockgroups returns a Vec of blockgroups instead of a single blockgroup since now we're in a multiple-bg realm
* Refactored the merge behavior again to duplicate parent -> child blockgroups/accessions/path instead of merge

In using it, a few other pieces were bolted on:

* Updated the view page to show sample hierarchy
* Added a sample level get_all_sequences for testing mostly
* Added default values for the sample name + parent names to keep the cli simpler in basic cases

### DIFF
--- a/gen-core/Cargo.toml
+++ b/gen-core/Cargo.toml
@@ -21,7 +21,7 @@ serde = {  version = "1.0.219", features = ["derive"] }
 thiserror = "1.0"
 sha2 = "0.10.9"
 hex = "0.4.3"
-rand = "0.9.2"
+rand = "0.10.1"
 uuid = { version = "1.18.1", features = ["v7"] }
 pyo3 = { version = "0.25.1", optional = true }
 pyo3-macros = { version = "0.25.1", optional = true }
--- a/gen-models/src/block_group.rs
+++ b/gen-models/src/block_group.rs
@@ -79,17 +79,25 @@ impl<'a> Capnp<'a> for BlockGroup {
         let sample_name = reader.get_sample_name().unwrap().to_string().unwrap();
         let name = reader.get_name().unwrap().to_string().unwrap();
         let created_on = reader.get_created_on();
-        let parent_block_group_id = match reader.get_parent_block_group_id().which().unwrap() {
-            block_group::parent_block_group_id::None(()) => None,
-            block_group::parent_block_group_id::Some(parent_block_group_id) => Some(
-                parent_block_group_id
-                    .unwrap()
-                    .as_slice()
-                    .unwrap()
-                    .try_into()
-                    .unwrap(),
-            ),
-        };
+        let parent_block_group_id =
+            reader
+                .get_parent_block_group_id()
+                .which()
+                .ok()
+                .and_then(|parent_block_group_id| match parent_block_group_id {
+                    block_group::parent_block_group_id::None(()) => None,
+                    block_group::parent_block_group_id::Some(parent_block_group_id) => {
+                        if let Ok(parent_block_group_id) = parent_block_group_id {
+                            if let Some(parent_block_group_id) = parent_block_group_id.as_slice() {
+                                parent_block_group_id.try_into().ok()
+                            } else {
+                                None
+                            }
+                        } else {
+                            None
+                        }
+                    }
+                });
         let is_default = reader.get_is_default();
 
         BlockGroup {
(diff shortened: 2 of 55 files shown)


gen-models/src/sequence.rs
```diff
@@ -394,7 +394,7 @@ mod tests {
     use std::time;
     use std::{fs::OpenOptions, io::Write};
 
-    use rand::{self, Rng};
+    use rand::{self, RngExt};
```
Chris7, line 397: had to upgrade rand for cargo deny

gen-core/Cargo.toml
```diff
@@ -21,7 +21,7 @@ serde = {  version = "1.0.219", features = ["derive"] }
 thiserror = "1.0"
 sha2 = "0.10.9"
 hex = "0.4.3"
-rand = "0.9.2"
```
Chris7, line 24: had to upgrade rand for cargo deny

gen-models/src/block_group.rs
```diff
@@ -82,6 +119,15 @@ pub struct BlockGroupData<'a> {
     pub name: String,
 }
 
+#[derive(Clone, Debug, Default)]
+pub struct NewBlockGroup<'a> {
```
Chris7, line 123: updated the create signature to take a struct so we can expand simpler in the future

gen-models/src/block_group.rs
```diff
@@ -1358,51 +1480,115 @@ mod tests {
...
         assert_eq!(
-            all_sequences,
+            BlockGroup::get_all_sequences(conn, &child_a.id, false),
+            HashSet::from_iter(vec!["AAAA".to_string()])
+        );
+        assert_eq!(
+            BlockGroup::get_all_sequences(conn, &child_b.id, false),
+            HashSet::from_iter(vec!["CCCC".to_string()])
+        );
+        assert_eq!(
```
Chris7, line 1540: This little bit of test made this entire PR happen

src/exports/genbank.rs
```diff
@@ -228,10 +228,10 @@ pub fn export_genbank(
                                     .splice(upos..upos, sequence.into_bytes())
                                     .collect::<Vec<_>>();
                                 qualifiers.push((
-                                    QualifierKey::from("note"),
+                                    Cow::Borrowed("note"),
```
Chris7, line 231: gb-io had a cargo deny thing too. This is the new api

src/updates/fasta.rs
```diff
@@ -48,138 +48,149 @@ pub fn update_with_fasta(
...
+            }
+        })
+        .collect::<Vec<_>>();
 
-    // Assuming just one entry in the fasta file
-    let mut first_node = None;
     let mut change_count = 0;
     for (index, result) in fasta_reader.records().enumerate() {
         let record = result?;
-
```
Chris7, line 69: this stuff was nested one level. So we iterate over the fasta, then iterate over each BG, then do the change. I did this instead of loading the entire fasta into memory.

src/updates/vcf.rs
```diff
@@ -340,38 +345,42 @@ pub fn update_with_vcf(
...
-                        let path_length = path_lengths
-                            .entry(sample_path.id)
-                            .or_insert_with(|| sample_path.length(conn));
-
-                        if ref_start > *path_length {
-                            return Err(VcfError::InvalidRecord(format!(
-                                "Invalid position found. Path {0} has length of {path_length}, change is in position {ref_start}.",
-                                sample_path.name
-                            )));
+                        for sample_bg_id in &sample_bg_ids {
```
Chris7, line 348: handling the multi-bg case

src/views/block_group.rs
```diff
@@ -202,7 +202,7 @@ pub fn view_block_group(
         CollectionExplorer::new(conn, op_conn, sample_name.as_deref(), collection_name);
     let mut explorer_state = CollectionExplorerState::new();
     if let Some(ref s) = sample_name {
-        explorer_state.toggle_sample(s);
+        explorer_state.set_sample_expanded(s, true);
```
Chris7, line 205: visual changes for this and few files below to make samples its own widget instead of part of the collection panel, and have the nested view

gen-models/src/block_group.rs
dkhofer: :+1:

gen-core/Cargo.toml
dkhofer: :+1: Hope my PR helps

gen-models/src/sequence.rs
dkhofer: :+1:

src/exports/genbank.rs
dkhofer: Weird flex but OK

src/updates/fasta.rs
dkhofer: :+1:

src/updates/vcf.rs
dkhofer: :+1:

src/views/block_group.rs
dkhofer: :+1:

dkhofer (approved):
Overall seems reasonable.  LGTM

PR #93, opened 2026-04-14
GFA: derive block group name from filename on import

### LINKED ISSUE #92: GFA import - block group naming
### Problem
When importing a GFA file (via CLI or Python bindings), the sequence graph (aka block group) is always created with an **empty name** (`""`), unlike FASTA and GenBank imports which use the entry name from the file.
**Comparison:**
| Import Type | CLI Block Group Name | Python Block Group Name |
|-------------|---------------------|-------------------------|
| GFA | Always `""` | Always `""` |
| FASTA | From file (sequence name/ID) | From file |
| GenBank | From file (locus name) | From file |
| Library | CLI arg `library_name` | Arg `library_name` |

**Relevant code locations:**
- CLI: `src/commands/import/gfa.rs:52` - passes `""` to `BlockGroup::create`
- Python: `gen-python/src/imports.rs:80` - calls CLI function with empty name

### Impact
This affects both CLI and Python users:
- All GFA-imported block groups have the same (empty) name
- Difficult to identify or query specific GFA imports afterward
- Inconsistent behavior compared to other import types
- Importing the same GFA file multiple times keeps adding more nodes and edges to the same object

### Short-term Solution
Use the **filename without extension** as the default block group name, with an optional parameter to override it.
**CLI changes** (`src/commands/import/gfa.rs`):
1. Add new CLI argument: `--graph-name` (optional) to explicitly set the block group name
2. If not provided, derive from filename: `Path::new(&path).file_stem().to_string_lossy()`
**Python bindings** (`gen-python/src/imports.rs`):
1. Add optional `graph_name: Option<String>` parameter to `import_gfa()`
2. Pass through to Rust function; derive from filename if not provided

### Example Usage
**CLI:**
```bash
# Uses "my_design" as block group name (derived from filename)
gen import gfa my_design.gfa --sample s1
# Explicitly overrides block group name to "custom_name"
gen import gfa my_design.gfa --sample s1 --graph-name custom_name
```

Python:
```python
# Uses "my_design" as block group name (derived from filename)
db.import_gfa("my_design.gfa", sample="s1")
# Explicitly overrides block group name to "custom_name"
db.import_gfa("my_design.gfa", sample="s1", graph_name="custom_name")
```

### Related: Larger model renaming effort 
The current --name flag for collection is confusing. What I'm hoping we conclude there will have us treating collections like filesystem paths with analogous navigation concepts (nesting, cd, pwd, ls). Reserving --name for the sequence graph name seems sensible, provided we have a good way to deal with multiple contigs per file.

bobvh:
GFA imports always created block groups with an empty name, making them indistinguishable when importing multiple files and preventing correct deduplication checks.

- Add `block_group_name` param to core `import_gfa`
- CLI: add `--graph-name` flag; defaults to file stem
- Python: add optional `graph_name` kwarg; defaults to file stem
- All internal/test callers pass "" to preserve existing behavior

### DIFF
--- a/src/imports/gfa.rs
+++ b/src/imports/gfa.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::HashMap,
+    collections::{HashMap, HashSet},
     fs::File,
     io::{BufRead, BufReader},
     path::Path as FilePath,
@@ -9,7 +9,6 @@ use gen_core::{
     HashId, NO_CHROMOSOME_INDEX, PATH_END_NODE_ID, PATH_START_NODE_ID, Strand, calculate_hash,
     is_end_node, is_start_node,
 };
-use gen_graph::{GraphEdge, GraphNode};
 use gen_models::{
     block_group::{BlockGroup, NewBlockGroup},
     block_group_edge::{BlockGroupEdge, BlockGroupEdgeData},
@@ -31,7 +30,10 @@ use gen_models::{
 };
 use indexmap::IndexSet;
 use itertools::Itertools;
-use petgraph::{algo::kosaraju_scc, prelude::UnGraphMap, visit::Dfs};
+use petgraph::{
+    prelude::UnGraphMap,
+    visit::{Bfs, Dfs},
+};
 use thiserror::Error;
 
 use crate::{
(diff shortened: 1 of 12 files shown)


Chris7:
Something to consider is a whole genome and multiple chromosomes in a GFA. I think this is how to do GFA imports:

Create a graph of the nodes/segments. Identify subgraphs and that is a blockgroup. If there exists a ref tag, use it, if not use the longest path name if they exist as the name. If neither exists, but there is only a single graph, give it a name based on the file.

IF  none of those conditions are met and there are multiple blockgroups, throw them all under the filename in a blockgroup with multiple graphs. Ensure to link the graphs into start/end nodes.

dkhofer:
For all the tests that call import_gfa, it'd be good to provide a string like "test graph" or "test sequence graph" instead of the empty string for the graph name

src/imports/gfa.rs
```diff
@@ -79,21 +84,71 @@ pub fn import_gfa(
...
-        },
-    )?;
+
     let bar = progress_bar.add(get_time_elapsed_bar());
     bar.set_message("Parsing GFA");
     let gfa: Gfa<String, (), ()> = Gfa::parse_gfa_file(gfa_path.to_str().unwrap());
+    let sn_tags = read_sn_tags(gfa_path);
+    bar.finish();
+
+    // Determine block group name per segment: SN:Z: tags > longest path/walk > filename stem
```
bobvh: This addresses @Chris7 ' suggestion

src/imports/gfa.rs
```diff
@@ -104,14 +159,29 @@ pub fn import_gfa(
             .sequence(input_sequence)
             .save(conn)?;
         sequences_by_segment_id.insert(&segment.id, sequence.clone());
-        // TODO: Node hash is always new, it's sorted by insert time via being a v7 uuid but maybe want to
```
bobvh, line 108: This was making it so that importing the same GFA twice stacks the graphs on top of each other  so you get for a graph A-B
```
start - A - B - end
      \       /
       A' - B' 
```

src/imports/gfa.rs
```diff
@@ -468,7 +578,32 @@ pub fn import_gfa(
     )
     .map_err(GFAImportError::OperationError);
     gen_bar.finish();
-    op
+    op.map(|op| (op, block_groups))
+}
+
+fn read_sn_tags(gfa_path: &FilePath) -> HashMap<String, String> {
```
bobvh, line 617: Should we have first parsed the whole file instead of extracting like this?

bobvh:
> Something to consider is a whole genome and multiple chromosomes in a GFA. I think this is how to do GFA imports:
> 
> Create a graph of the nodes/segments. Identify subgraphs and that is a blockgroup. If there exists a ref tag, use it, if not use the longest path name if they exist as the name. If neither exists, but there is only a single graph, give it a name based on the file.
> 
> IF none of those conditions are met and there are multiple blockgroups, throw them all under the filename in a blockgroup with multiple graphs. Ensure to link the graphs into start/end nodes.

On rereading there's an issue with my implementation: connected segments with different SN tags are split between bgs, and disconnected subgraphs with the same name have the opposite issue. Working on a fix.

src/imports/gfa.rs
```diff
@@ -1109,12 +1151,42 @@ mod tests {
         track_database(conn, op_conn).unwrap();
         let _ = import_gfa(&context, &gfa_path, &collection_name, Sample::DEFAULT_NAME);
 
+        let block_groups = BlockGroup::query(
```
bobvh, line 1154: todo: replace sql query with proper method call

PR #94, opened 2026-04-14
Add graph search algorithms with CLI exposure

bobvh:
Implements exact-match sequence search over the graph using a BFS-based matcher (GenGraphMatcher) and an optional k-mer seed index (SeedIndex) for fast lookup. Exposes three new CLI subcommands: `build-index`, `clear-index`, and `search`. Adds postcard as a serialization dependency for persisting the seed index to disk.


src/graphs/graph_search.rs
```diff
@@ -0,0 +1,1333 @@
...
+
+/// Bumped whenever the index format or indexing behavior changes incompatibly.
+const SEED_INDEX_VERSION: u32 = 1;
+
+/// File header written before the `SeedIndex` payload.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+struct SeedIndexHeader {
+    version: u32,
+    k: usize,
+    case_sensitive: bool,
```
bobvh: A flag to set case sensitivity hasn't been implemented yet. I'm not sure what the default should be, but I definitely already ran into problems myself from not realizing I had a mix of cases in my block groups.

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,1333 @@
...
+        assert_eq!(minus_35_hits.len(), 1);
+        assert_eq!(minus_35_hits[0].nodes.len(), 5); // five single-byte nodes
+
+        // ── 6. Absent sequence ───────────────────────────────────────────────
+        // No node "A" is immediately downstream of segment 11; only 12 ("T")
+        // and 13 ("C") are.
+        assert!(!matcher.contains(b"GCTAGCTCAGA"));
+
+        // ── 7. Lowercase does NOT match the Rust matcher ─────────────────────
+        // The Python `repo.search()` API normalises the query to uppercase before
```
bobvh: Python bindings are in a later commit. This is subject to change.

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,1333 @@
...
+    version: u32,
+    k: usize,
+    case_sensitive: bool,
+}
+
+/// Errors from `SeedIndex` serialization/deserialization.
+#[derive(Debug)]
+pub enum SeedIndexIoError {
+    Io(std::io::Error),
+    Encode(postcard::Error),
```
bobvh, line 583: I picked postcard as a simple binary serializer that works with serde. Is this something you would use protobuf or captnproto for?

src/commands/mod.rs
```diff
@@ -248,6 +248,47 @@ pub enum Commands {
...
+        #[arg(short, long)]
+        collection: Option<String>,
+        /// Restrict clearing to a specific sample
+        #[arg(short, long)]
+        sample: Option<String>,
+    },
+    /// Search for an exact sequence across all block groups
+    ///
+    /// Each match is reported with a blocks column formatted as [hash:start-end, ...],
+    /// where hash is a 12-character node hash prefix from the original node the block
```
Chris7, line 287: Why truncate it

bobvh:
Mostly because that’s how my git client and IDEs have been showing commit
hashes and I figured collision risk would be small enough at that length.
Did I get birthday-paradoxed?

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+        if query.is_empty() {
+            return true;
+        }
+
+        let matcher = self.sequence_kind.matcher_for_query(query);
+        if self.contains_query_orientation(query, matcher) {
+            return true;
+        }
+
+        if self.sequence_kind == SequenceKind::Dna {
```
Chris7, line 270: Why not make the matcher sensitive to rc vs doing the search twice

bobvh: Because the k-mer index was set up in the forward direction. It's probably not that much work to allow it to also search backwards. Or have two different matchers for with and without index. If we end up being happy with the performance on full size pangenomes without index, and we drop the index, then it definitely makes sense to do it during matching.

src/commands/mod.rs
bobvh: see top level email reply

src/graphs/graph_search.rs
bobvh: this is resolved in later commit

src/graphs/graph_search.rs
bobvh: Resolved

gen-models/src/db.rs
```diff
@@ -58,6 +58,12 @@ impl<C> DbHandle<C> {
 pub type GraphHandle = DbHandle<GraphConnection>;
 pub type OperationsHandle = DbHandle<OperationsConnection>;
 
+impl DbHandle<GraphConnection> {
```
Chris7: i don't see this being used

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+/// `blocks.last()` is the last. `start_offset` and `end_offset` anchor the
+/// match within those boundary nodes.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct GraphLocus {
+    /// Local offset of the first matched byte inside `blocks[0]`.
+    pub start_offset: usize,
+    /// Local offset past the last consumed byte in `blocks.last()`.
+    pub end_offset: usize,
+    /// Ordered sequence of nodes that spell the match. Length is at least 1.
+    pub blocks: Vec<GraphNode>,
```
Chris7, line 36: i'm not sure how this is enough to match a path. Wouldn't we need to track the edges as well?

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+
+/// A linear walk through graph space that covers a complete match.
+///
+/// `blocks[0]` is the first node containing matched bytes;
+/// `blocks.last()` is the last. `start_offset` and `end_offset` anchor the
+/// match within those boundary nodes.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct GraphLocus {
+    /// Local offset of the first matched byte inside `blocks[0]`.
+    pub start_offset: usize,
```
Chris7, line 32: if you tracked GraphPos instead is this needed?

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+    /// bases; otherwise uses normalized exact matching.
+    SsDna,
+    /// Protein or other non-nucleotide sequence search.
+    ///
+    /// Case-insensitive (both sides uppercased). No IUPAC expansion or reverse
+    /// complement.
+    Protein,
+}
+
+impl Default for SequenceKind {
```
Chris7: btw you can use #[default] in the above enum and not need this separate impl

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+    fn default() -> Self {
+        Self::Dna
+    }
+}
+
+impl SequenceKind {
+    /// Compatibility helper for call sites that still use string sequence types.
+    pub fn from_seq_type(seq_type: Option<&str>) -> Self {
+        match seq_type {
```
Chris7, line 66: can you use the eq_ignore_ascii_case here as well to just ignore case?

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+    /// bases. Returns an error if the query contains degenerate bases, because an
+    /// exact seed lookup would silently miss valid matches.
+    ///
+    /// Normalization of the seed lookup and full match follows `seed_index.normalized`.
+    pub fn find_all_with_seed_index(
+        &self,
+        seed_index: &SeedIndex,
+        query: &[u8],
+    ) -> Result<Vec<GraphLocus>, SeedIndexSearchError> {
+        if query.iter().any(|&byte| is_degenerate_iupac(byte)) {
```
Chris7, line 315: why don't we keep degenrates if the user is searching for them?

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+            return Ok(self.find_all_forward(query, matcher));
+        }
+
+        let seed: Vec<u8> = if seed_index.normalized {
+            query[..seed_index.k]
+                .iter()
+                .map(|&b| b.to_ascii_uppercase())
+                .collect()
+        } else {
+            query[..seed_index.k].to_vec()
```
Chris7, line 339: one place you uppercase the query, the other you don't. It should be case insenstive on both paths.

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+            if state.q_idx == query.len() {
+                return true;
+            }
+
+            if dead.contains(&state) {
+                continue;
+            }
+
+            match self.step(state, query[state.q_idx], matcher) {
+                StepResult::Advance(next) => stack.push(next),
```
Chris7, line 402: is there a reason to distinguish between branches and non-branches? Seems like the branch case covers it all.

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+            ),
+        }
+    }
+}
+
+impl std::error::Error for SeedIndexSearchError {}
+
+/// Errors from `SeedIndex` serialization/deserialization.
+#[derive(Debug)]
+pub enum SeedIndexIoError {
```
Chris7, line 579: we use thiserror for defining errors.

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+        expected: u32,
+    },
+    /// Header `k` does not match the requested k-mer length.
+    KMismatch {
+        got: usize,
+        expected: usize,
+    },
+}
+
+impl std::fmt::Display for SeedIndexIoError {
```
Chris7: use thiserror instead of this

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+impl From<std::io::Error> for SeedIndexIoError {
+    fn from(e: std::io::Error) -> Self {
+        Self::Io(e)
+    }
+}
+
+impl SeedIndex {
+    /// Build a k-mer index over the graph.
+    ///
+    /// When `normalized` is `true`, k-mer bytes are uppercased before insertion
```
Chris7, line 600: instead of messing with case, use something that has case-insensitive comparisons

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+
+    /// Serialize `self` to bytes: 4-byte little-endian header length, header, payload.
+    pub fn to_bytes_with_header(&self) -> Result<Vec<u8>, SeedIndexIoError> {
+        let header = SeedIndexHeader {
+            version: SEED_INDEX_VERSION,
+            k: self.k,
+            normalized: self.normalized,
+        };
+
+        let header_bytes = postcard::to_allocvec(&header).map_err(SeedIndexIoError::Encode)?;
```
Chris7, line 627: why postcard?

src/graphs/graph_search_tests.rs
```diff
@@ -0,0 +1,387 @@
+use gen_models::{block_group::BlockGroup, collection::Collection};
```
Chris7: can you move these to graph_search.rs at the end. That's the pattern we follow elsewhere atm

src/main.rs
```diff
@@ -618,6 +640,100 @@ fn call_cli() -> Result<(), Box<dyn std::error::Error>> {
...
+        }) => {
+            let collection_name = match collection {
+                Some(c) => c,
+                None => get_default_collection(operation_conn)?,
+            };
+            let block_groups = match sample {
+                Some(ref s) => Sample::get_block_groups(graph_conn, &collection_name, s),
+                None => Collection::get_block_groups(graph_conn, &collection_name),
+            };
+            let index_dir = workspace
```
Chris7, line 641: Should make the index_dir call part of the Workspace struct so it's centralized there.

src/main.rs
```diff
@@ -54,6 +57,25 @@ use itertools::Itertools;
 use rusqlite::{Connection, params, types::Value};
 use sha2::digest::typenum::Gr;
 
+// Format a match's path as "[hash:start-end, ...]" with 12-char node hash prefixes.
+// start/end are the node's own coordinates in the reference sequence.
+fn fmt_match_path(m: &GraphLocus) -> String {
```
Chris7: this shouldn't be part of the parser interface

Chris7: i see it's just used in one place too, just put it inline there.

src/main.rs
```diff
@@ -618,6 +640,100 @@ fn call_cli() -> Result<(), Box<dyn std::error::Error>> {
...
+            let block_groups = match (collection, sample.as_deref()) {
+                (Some(c), Some(s)) => Sample::get_block_groups(graph_conn, &c, s),
+                (Some(c), None) => Collection::get_block_groups(graph_conn, &c),
+                (None, Some(s)) => BlockGroup::all(graph_conn)
+                    .into_iter()
+                    .filter(|bg| bg.sample_name == s)
+                    .collect(),
+                (None, None) => BlockGroup::all(graph_conn),
+            };
+            let index_dir = workspace.find_gen_dir().map(|d| d.join("search_index"));
```
Chris7: add this to workspace, find_search_index or something

src/main.rs
```diff
@@ -618,6 +640,100 @@ fn call_cli() -> Result<(), Box<dyn std::error::Error>> {
...
+                let graph = BlockGroup::get_graph(graph_conn, &bg.id);
+                let matcher = GenGraphMatcher::new(graph_conn, graph);
+                let matches = index_dir
+                    .as_ref()
+                    .and_then(|dir| {
+                        let p = dir.join(format!("{}.bin", bg.id));
+                        SeedIndex::load_from_path(p).ok()
+                    })
+                    .map(|idx| matcher.find_all_with_seed_index(&idx, query_bytes))
+                    .unwrap_or_else(|| Ok(matcher.find_all(query_bytes)))?;
```
Chris7, line 706: the presence of an index changes the outcomes. If an index is present, it looks for seed indices which is forward-only. find_all does RC as well.

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+            return Vec::new();
+        }
+
+        let matcher = self.sequence_kind.matcher_for_query(query);
+        let mut out = self.find_all_query_orientation(query, matcher);
+
+        if self.sequence_kind == SequenceKind::Dna {
+            let rc = reverse_complement(query);
+            let rc_matcher = self.sequence_kind.matcher_for_query(&rc);
+            out.extend(self.find_all_query_orientation(&rc, rc_matcher));
```
Chris7, line 298: if looking for palindromes, this would append the matching sequence twice since the forward was added above.

gen-models/src/db.rs
bobvh: You're right. Iirc at one point the AI had messed with the way we do tests and I must have not reverted it completely. Thank you for catching this, I should definitely have stopped it since it's in gen-models. Deleted.

src/graphs/graph_search.rs
bobvh: No that's the key part I've come to appreciate: blocks contain their start and end coordinate, so if you put them in an ordered list you have all the edges except first and last. To then narrow it down to a path that doesn't have to align to the sides of a variant, two offsets are enough. The reason for doing this in "block space" here rather than the DB layer nodes, is that it gets more intuitive for search and plots.

src/graphs/graph_search.rs
bobvh: It's just because GraphPos also includes a block itself. So I'd have either the same block stored twice (once in the list, once in the GraphPos), or would have to keep the list of blocks minus the first and last, which would make for more annoying loops. No strong opinion on it though.

src/graphs/graph_search.rs
bobvh: TIL, thanks

src/commands/mod.rs
bobvh: Mostly because that’s how my git client and IDEs have been showing commit
hashes and I figured collision risk would be small enough at that length.
Did I get birthday-paradoxed?

src/graphs/graph_search.rs
bobvh: I'll consolidate one doing everything with the eq_ignore instead of comparing after uppercasing/lowercasing

src/graphs/graph_search.rs
bobvh: The index doesn't support degenerate matching. Or are you saying: the user actually wants to find NNNACTG literally?

src/graphs/graph_search.rs
bobvh: Will fix with eq_ignore_ascii_case

src/graphs/graph_search.rs
bobvh: I picked postcard as a simple binary serializer that works with serde. Is this something you would use protobuf or captnproto for?

src/main.rs
bobvh: Not if I perform the RC match by doing the search a second time on RC(query), no?

src/graphs/graph_search.rs
bobvh: That's technically correct though: one match has strand +, one match has strand -

src/graphs/graph_search.rs
bobvh: You mean: why not treat advancing within a node as a junction with 1 outcome? I suppose we could. You'd be allocating a million Vec vs a million bytes for a large reference genome though, wouldn't that be a lot of overhead?

src/graphs/graph_search.rs
bobvh: Fixed

src/graphs/graph_search.rs
bobvh: Fixed

src/graphs/graph_search_tests.rs
bobvh: I had m like that too initially but thought the file was getting too long (>1k lines with tests). Moved.

src/main.rs
bobvh: done

src/main.rs
bobvh: done

src/main.rs
bobvh: done

src/graphs/graph_search.rs
bobvh: (I ended up not eliminating it everywhere because the keys on the main hashmap still need to be matched exactly, but it's consistent where I do uppercase)

src/graphs/graph_search.rs
bobvh: Or as an alternative, introduce NodeSpan (node_id, start, end) . It matches how you did AnnotationSegment iirc. Similar idea: every block is a NodeSpan, not every NodeSpan is a block.

That way GraphLocus becomes just a Vec<NodeSpan> and our edges are derived from the order in that list.

I'd also drop GraphPos (relative to block start) then in favor of something like NodePos (relative to the node), so we can consider edges as (NodePos, NodePos).

(Separate PR though)

src/graphs/graph_search.rs
bobvh: See also this point: https://github.com/genhub-bio/gen/pull/94/changes/BASE..b95f8860cfec3260938c5944c2d43fd63e6d885e#r3175829576

src/graphs/graph_search.rs
```diff
@@ -0,0 +1,815 @@
...
+
+/// A position in the graph: a GraphNode (aka Block) plus a local byte offset
+/// relative to the start of that GraphNode's sequence slice.
+///
+/// If the GraphNode has a sequence_start of 100, then offset 5 means sequence
+/// position 105 in the underlying stored sequence.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct GraphPos {
+    pub node: GraphNode,
+    /// Local offset within contained sequence slice: `0..=node.length()`.
```
bobvh: I mixed metaphors here, to be correct with the GraphPos docstring and usage this needs to be GraphNode.length or block.length

gen-models/src/db.rs
bobvh: Update: it was for the Python bindings. Blockgroups used it to get the repository they came from. The tests didn't catch it because we didn't have the coverage. I'll fix it.

PR #96, opened 2026-04-14
Upgrade crates

dkhofer:


dkhofer:
Hoping this resolves all the security alerts: https://github.com/genhub-bio/gen/security/dependabot

PR #97, opened 2026-04-16
Fix diff operation on empty branch

Chris7:
If the from_hash is None, we should do all the upstream operations of to_hash, not just to_hash. When doing a diff against an empty branch, we currently just return the diff of the tip of the source branch instead of all operations.


dkhofer (approved):
LGTM

PR #101, opened 2026-04-21
Query poc

Chris7:
Adds a query builder macro, so we can take somtehing like

```
#[derive(ModelSelect)]
struct Sample {
  name: String
}
```

and be able to use it like `Sample::select(conn).name_contains("foo")`. Supports joins/limits as well. I need this because other applications built on the library want to query/load data in ways our public api doesn't support. This seems like a good way to limit and control the public surface area of how our models can be queried without getting into the mess that techniques like reflection have.

I used this for a few models to show how it can be used.


gen-models/src/block_group.rs
```diff
@@ -1156,7 +1157,8 @@ impl BlockGroup {
     ) -> Result<Path, BlockGroupError> {
         let query = format!(
             "SELECT id, block_group_id, name, created_on FROM {} \
-             WHERE block_group_id = :block_group_id ORDER BY created_on DESC",
+             WHERE block_group_id = :block_group_id \
```
Chris7: it decided to optimize this query

gen-models/src/select.rs
```diff
@@ -0,0 +1,446 @@
+//! Internal support for selectors generated by `ModelSelect`.
```
Chris7, line 1: select helpers

gen-models-macros/src/lib.rs
```diff
@@ -0,0 +1,534 @@
+//! Derive support for typed, fluent SQL selectors in `gen-models`.
```
Chris7, line 1: This adds a macro for our structs that generates query builder helpers

dkhofer: 👍

gen-models-macros/src/lib.rs
```diff
@@ -0,0 +1,590 @@
...
+/// aliased queries.
+#[proc_macro_derive(ModelSelect, attributes(model_select))]
+pub fn derive_model_select(input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    expand_model_select(input)
+        .unwrap_or_else(Error::into_compile_error)
+        .into()
+}
+
+fn expand_model_select(input: DeriveInput) -> syn::Result<proc_macro2::TokenStream> {
```
dkhofer, line 36: This needs a significant comment explaining the context of what it's doing

gen-models-macros/src/lib.rs
```diff
@@ -0,0 +1,590 @@
...
+        .as_ref()
+        .map(|alias| {
+            let alias = LitStr::new(alias, Span::call_site());
+            quote! { #alias }
+        })
+        .unwrap_or_else(|| {
+            quote! { <#model as ::gen_models::traits::Query>::TABLE_NAME }
+        });
+
+    let mut variants = Vec::new();
```
dkhofer: Maybe this should be called "selectable_fields"?  "variants" is ambiguous from the biological context, too

gen-models-macros/src/lib.rs
```diff
@@ -0,0 +1,590 @@
+//! Derive support for typed, fluent SQL selectors in `gen-models`.
+
+use proc_macro::TokenStream;
+use proc_macro2::Span;
+use quote::{format_ident, quote};
+use syn::{
+    Attribute, Data, DeriveInput, Error, ExprPath, Fields, Ident, LitStr, Type, parse_macro_input,
+};
+
+/// Generates a `<Model>Select` query builder from a persisted model's named fields.
```
dkhofer: I would reword this a fair amount.  Something like

```
When added as a trait to a gen model class, ModelSelect gives the ability to build and then create a sql statement by invoking method calls like `select()`, `order()`, and `join()` on the model.  We parse the model's named fields so they are available to the methods used to build the sql query.
```

gen-models-macros/src/lib.rs
dkhofer: Like, I get that DeriveInput is some kind of parsed version of the model class's information, but I'm not following how or why all the DeriveInput object's fields line up to things used in a sql query

gen-models-macros/src/lib.rs
```diff
@@ -0,0 +1,590 @@
...
+    if !input.generics.params.is_empty() {
+        return Err(Error::new_spanned(
+            input.generics,
+            "ModelSelect does not support generic model structs",
+        ));
+    }
+
+    let options = ContainerOptions::from_attributes(&input.attrs)?;
+    let selector = format_ident!("{}Select", model);
+    let source_alias = options
```
dkhofer, line 79: What is an alias here?

dkhofer: Like, what in a rust struct defines an alias?

gen-models-macros/src/lib.rs
```diff
@@ -0,0 +1,590 @@
...
+            }
+
+            fn column(&self, column: &str) -> ::std::string::String {
+                ::gen_models::select::qualify_sql_column(
+                    <Self as ::gen_models::select::SelectQuery>::source(self).alias(),
+                    column,
+                )
+            }
+
+            pub fn with_ref(mut self, history_ref: impl ::core::convert::Into<::std::string::String>) -> Self {
```
dkhofer: Could you add a comment explaining this is a dolt thing?

gen-models/src/block_group.rs
```diff
@@ -1156,7 +1159,8 @@ impl BlockGroup {
...
```
dkhofer: This doesn't seem related to the query functionality?  Why is it changing?  Is it not possible to construct this with the new machinery?

gen-models/src/select.rs
dkhofer: OK

gen-models/src/select.rs
```diff
@@ -0,0 +1,809 @@
...
+impl_select_projection_for_tuple!(A:0, B:1, C:2, D:3, E:4, F:5, G:6);
+impl_select_projection_for_tuple!(A:0, B:1, C:2, D:3, E:4, F:5, G:6, H:7);
+impl_select_projection_for_tuple!(A:0, B:1, C:2, D:3, E:4, F:5, G:6, H:7, I:8);
+impl_select_projection_for_tuple!(A:0, B:1, C:2, D:3, E:4, F:5, G:6, H:7, I:8, J:9);
+impl_select_projection_for_tuple!(A:0, B:1, C:2, D:3, E:4, F:5, G:6, H:7, I:8, J:9, K:10);
+impl_select_projection_for_tuple!(A:0, B:1, C:2, D:3, E:4, F:5, G:6, H:7, I:8, J:9, K:10, L:11);
+impl_select_projection_for_tuple!(A:0, B:1, C:2, D:3, E:4, F:5, G:6, H:7, I:8, J:9, K:10, L:11, M:12);
+impl_select_projection_for_tuple!(A:0, B:1, C:2, D:3, E:4, F:5, G:6, H:7, I:8, J:9, K:10, L:11, M:12, N:13);
+impl_select_projection_for_tuple!(A:0, B:1, C:2, D:3, E:4, F:5, G:6, H:7, I:8, J:9, K:10, L:11, M:12, N:13, O:14);
+impl_select_projection_for_tuple!(A:0, B:1, C:2, D:3, E:4, F:5, G:6, H:7, I:8, J:9, K:10, L:11, M:12, N:13, O:14, P:15);
```
dkhofer, line 526: I believe the Rails code goes up to 42 for this

dkhofer: Actually, no, I think it's for methods like "once", "twice", etc.

dkhofer: OK, it's this: https://www.quora.com/Why-is-Array-forty_two-called-the-reddit-in-Ruby-on-Rails

dkhofer (approved):
OK, this is pretty cool.  You're basically reinventing Rails ActiveRecord, but it's not a crazy amount of code.

That said, are you sure there isn't a rust crate that already does this?  This is kind of outside of our line of business.

If there isn't such a crate, then to satisfy my concerns about feature creep, could you go through all the sql in our models and see how much of it can be converted to use the query builder?  (Or at least for the ones we care about for genhub's purposes.)  How much more functionality would need to be added to it to support all our use cases?  Can we keep the model select crate relatively small?

I do feel the macro for query building needs to be commented more, so I left a few comments there.

That said, this all basically seems fine, so I'm approving.

PR #102, opened 2026-04-21
R bindings

Chris7:


gen-r/README.md
```diff
@@ -0,0 +1,74 @@
...
+- `import_fasta()` imports sequence data into a collection.
+- `export_fasta()` exports a collection or sample back to FASTA.
+
+## Installation
+
+Install a prebuilt package from a tagged GitHub release. This is the supported
+end-user path on macOS and Windows and does not require Rust, cargo,
+`libclang`, or `capnp`.
+
+On macOS and Windows, install the published R package binary:
```
bobvh, line 30: The R addon artefact isn't in releases currently (for example: https://github.com/genhub-bio/gen/releases/download/v0.1.31/genr_0.1.31-macos-arm64.tgz )

gen-r/README.md
```diff
@@ -0,0 +1,74 @@
...
+  version,
+  asset
+))
+```
+
+Linux installs currently require a source build. For development or Linux,
+install Rust/cargo, `libclang`, and `capnp`, then use:
+
+```r
+remotes::install_github("genhub-bio/gen", subdir = "gen-r", ref = "v0.1.31")
```
bobvh, line 60: 404 error for me

`Failed to install 'unknown package' from GitHub:
  cannot open URL 'https://api.github.com/repos/genhub-bio/gen/contents/gen-r/DESCRIPTION?ref=v0.1.31'`

gen-r/src/rust/src/lib.rs
```diff
@@ -0,0 +1,1429 @@
...
+    let node_id = hash_id_from_string(&node_id).map_err(Error::Other)?;
+    let sequences = Node::get_sequences_by_node_ids(&conn, &[node_id]);
+    let sequence = sequences
+        .get(&node_id)
+        .ok_or_else(|| Error::Other(format!("Node with id {node_id} not found")))?;
+    Ok(sequence.get_sequence(sequence_start, sequence_end))
+}
+
+#[extendr]
+fn graph_render_frame(
```
bobvh, line 1351: You really went for feature parity to Python, very cool.

gen-r/DESCRIPTION
```diff
@@ -0,0 +1,23 @@
+Package: genr
+Title: R Bindings for the Gen Sequence Graph Database
+Version: 0.1.31
+Authors@R:
+    person("GenHub", "Bio", email = "opensource@genhub.bio", role = c("aut", "cre"))
```
bobvh, line 5: Did you already set up that email alias?

gen-r/R/high_level.R
bobvh, line 1: Am I correct in understanding that this is basically the main file to edit going forward? (+ documentation in the man folder

gen-r/src/rust/src/lib.rs
bobvh: I haven't been able to make it work though. I get 
> "Error in graph_render_frame(ctrl$db_path, ctrl$block_group_id, ctrl$detail,  : 
>   Lazy instance has previously been poisoned"

I think that may come from the (old) color theme, I'm not sure if it goes away once rebased onto the simplified theme infra.

bobvh (commented):
I had to install it manually through a file that the public doesn't have access to. I haven't tested the full range of features in R, but the basics definitely work. Highest impact edits would be the installation, and I don't think we need full parity to the python bindings to release (we could comment out the bigger untested parts)

gen-r/DESCRIPTION
Chris7: no

gen-r/R/high_level.R
Chris7: maybe, this is a first attempt so there's nothing really beyond "it was a working approach"

gen-r/README.md
Chris7: there's no deployment yet so it wont work

gen-r/README.md
Chris7: it's in a branch so it won't be on main. Those come from main

PR #103, opened 2026-04-22
Fix config

Chris7:
Remove the local use of manifest_path which is not present in distributions. Renames to theme


Cargo.toml
```diff
@@ -7,6 +7,7 @@ repository = "https://github.com/genhub-bio/gen"
 homepage = "https://genhub.bio"
 license = "Apache-2.0"
 include = [
+  "/config",
```
Chris7: include the asset path

bobvh (approved):
I've simplified the theme infra in the branch I'm currently working in (no yaml dependencies, not passing the theme around everywhere) and placed the theme definitions in a /themes directory as json files. So that's compatible with what you're doing here.

PR #104, opened 2026-04-23
Genbank annotation export

Chris7:
Adds annotations to the genbank export if present.

### DIFF
--- a/gen-annotations/src/projection.rs
+++ b/gen-annotations/src/projection.rs
@@ -81,50 +81,6 @@ pub fn merge_annotation_segments(segments: Vec<AnnotationSegment>) -> Vec<Annota
     )
 }
 
-pub fn project_single_annotation_segment(
-    segment: &AnnotationSegment,
-    path_blocks: &[PathBlock],
-) -> Vec<AnnotationSegment> {
-    merge_annotation_segments(
-        path_blocks
-            .iter()
-            .filter_map(|block| {
-                if block.node_id != segment.node_id {
-                    return None;
-                }
-                let overlap_start = max(segment.range.start, block.sequence_start);
-                let overlap_end = min(segment.range.end, block.sequence_end);
-                if overlap_end <= overlap_start {
-                    return None;
-                }
-
-                let (start, end) = if block.strand == Strand::Reverse {
-                    (
-                        block.path_start + (block.sequence_end - overlap_end),
-                        block.path_start + (block.sequence_end - overlap_start),
-                    )
-                } else {
-                    (
-                        block.path_start + (overlap_start - block.sequence_start),
-                        block.path_start + (overlap_end - block.sequence_start),
-                    )
-                };
-                let strand = if block.strand == Strand::Reverse {
-                    segment.strand.complement()
-                } else {
-                    segment.strand
-                };
-
-                Some(AnnotationSegment {
-                    node_id: block.node_id,
-                    range: Range { start, end },
-                    strand,
-                })
-            })
-            .collect(),
-    )
-}
-
 pub fn project_annotation_segments(
     accession_segments: &[AnnotationSegment],
     path_blocks: &[PathBlock],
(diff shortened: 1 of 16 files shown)


gen-capnp-schemas/gen-models.capnp
```diff
@@ -253,6 +253,7 @@ struct Annotation {
   name @1 :Text;
   annotationGroup @2 :Text;
   accessionId @3 :List(UInt8);
+  extra @4 :Text;
```
Chris7, line 256: for storing extra stuff from file-based annotations if a user has imported them to the db.

gen-core/src/range.rs
```diff
@@ -197,52 +197,50 @@ impl Range {
...
+    merged
+}
+
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub struct RangeMapping {
     pub source_range: Range,
     pub target_range: Range,
 }
 
-impl RangeMapping {
```
Chris7, line 206: all the merging stuff/coordinate stuff felt wrong, then i remembered this. I extended the Range object to be mergable and handle the stuff I was going w/ merging anntoations.

gen-core/src/strand.rs
```diff
@@ -27,6 +27,15 @@ impl Strand {
         let b_ambig = Strand::is_ambiguous(b);
         (a_ambig || b_ambig) || a == b
     }
+
+    pub fn complement(self) -> Strand {
```
Chris7, line 31: surprised it took this long to need complement.

src/exports/genbank.rs
```diff
@@ -468,6 +692,302 @@ mod tests {
...
+            rusqlite::params![serde_json::to_string(&extra).unwrap(), annotation.id],
+        )
+        .unwrap();
+
+        let tmp_dir = tempfile::tempdir().unwrap().keep();
+        let filename = tmp_dir.join("out.gb");
+        export_genbank(conn, "test", "test", &filename).unwrap();
+
+        let exported_text = std::fs::read_to_string(&filename).unwrap();
+        assert!(
```
Chris7, line 977: these 2 assertions are working around the padded strings in genbanks and how gb-io parses them

src/imports/genbank.rs
```diff
@@ -6,7 +6,10 @@ use std::{
 };
```
Chris7, line 6: I feel like some of this file belongs in gen-annotations but i didn't want to add noise to this pr

src/exports/genbank.rs
```diff
@@ -122,7 +245,7 @@ pub fn export_genbank(
     conn: &GraphConnection,
     collection_name: &str,
     sample_name: &str,
-    filename: &PathBuf,
+    writer: impl Write,
```
Chris7, line 251: made this a writer so we don't need to path filenames and can stream to buffers

gen-capnp-schemas/gen-models.capnp
dkhofer: :+1:

gen-core/src/range.rs
dkhofer: :+1:

gen-core/src/strand.rs
dkhofer: Wow, yeah

src/exports/genbank.rs
dkhofer: 👍

src/exports/genbank.rs
dkhofer: 👍

src/imports/genbank.rs
dkhofer: 👍

gen-annotations/src/projection.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    segments
+}
+
+pub fn merge_annotation_segments(segments: Vec<AnnotationSegment>) -> Vec<AnnotationSegment> {
+    merge_ordered_items(
+        segments
+            .into_iter()
+            .filter(|segment| segment.range.end > segment.range.start)
```
dkhofer: This pattern of filtering for segment ranges with end > start shows up a lot here.  For my understanding, what does end < start mean, and why are we skipping ones like that?

dkhofer (approved):
All seems reasonable, LGTM

gen-annotations/src/projection.rs
Chris7: I'm going to remove it. It was meant to handle bad data, but i remember that end > start for origin ones so it needs to be supported.

PR #105, opened 2026-04-24
Improve annotation packing

Chris7:
This will pack more annotations into the tui view. Currently it just picks like 2/3 and will leave the rest of the row empty. This packs each row.


dkhofer:
Honestly I think Bob would be the better reviewer here, so I added him.  I'll approve so as not to hold things up, but it'd be good to get his feedback

dkhofer (approved):

bobvh:
Looks good to me.

PR #106, opened 2026-04-27
Simplify theme system

bobvh:
Consolidates global theme logic to a short file `src/base16.rs` (location up for discussion), removes the `color_utils` helper, and consolidates gen-tui colour-resolution logic into `gen-tui`'s `theme.rs`. Bundles two JSON theme files (`themes/latte.json`, `themes/mocha.json`) instead of yaml and wires `init_theme()` into `main()`. Cuts several colour-related crate dependencies. Colours are now referenced by their base16 standard IDs (0x00 to 0x0F) instead of layering on semantics.


Chris7 (approved):
Points I see:
* 
* Can you create the themes module and move base16 back?
* The new themes are in a bad folder, `themes`. That isn't included in the build, so it'll break. The old folder is `theme`. We should have a unit test or something b/c this has happened a few times w/ config files

Otherwise lgtm

bobvh:
All fixes made, thanks for the clear comment. The test added isn't going to catch trying to grab the themes without include_str , but it does test that it can find and parse them.

PR #108, opened 2026-04-27
Rewrite Jupyter widget in TypeScript and fix terminal artifacts

bobvh:
- **TypeScript rewrite** — replaces the vanilla JS bundle with TypeScript sources in `gen-python/js/`; compiled JS checked in so `npm` is not required at build time. Build via "make jupyter" in project root. interaction and zoom buttons.
- **Three-colour glyph rendering** — a terminal supports only 2 colours per cell, which produces artefacts when highlighting a path. But in the browser we can implement 3-colour rendering by splitting box-drawing characters into their heavy and light components, each painted a different colour. Avoids font-specific rendering issues and adjust character padding by intercepting certain characters during rendering (goal: better legibility).
- **Colour cycling and example notebooks** — auto-assigns highlight colours from the theme palette; adds `annotations.ipynb`, `yeast_brq_demo.ipynb`, `flo11_brq_demo.ipynb`, `comprehensive_api_demo.ipynb`, and reference data files. We shouldn't keep all of them, but condense a good showcase.


gen-python/python/gen/static/jupyter_widget.js.map
```diff
@@ -0,0 +1,7 @@
+{
```
Chris7, line 1: generally these won't be tracked by git. Is there a reason this one is?

gen-python/src/python_api/jupyter_widget.rs
```diff
@@ -268,46 +478,395 @@ impl PyGraphController {
...
+    pub fn show_path(&mut self, color: Option<&str>) -> PyResult<()> {
+        let block_group_id = self.block_group_id.ok_or_else(|| {
+            PyRuntimeError::new_err(
+                "show_path() requires a block group; obtain the widget via BlockGroup.plot()",
+            )
+        })?;
+
+        self.ensure_connection()?;
+
+        use ratatui::style::Color;
```
Chris7, line 533: ratatui installs in jupyter?

.gitignore
```diff
@@ -36,6 +36,7 @@ rust-analyzer.json
 # Python bindings
 **/__pycache__/
 gen-python/python/gen/*.so
+gen-python/node_modules/
```
Chris7, line 39: can you update the other node_modules to be **/node_modules/? Maybe it'll get them all then

Chris7 (approved):
Looks ok.

gen-python/python/gen/static/jupyter_widget.js.map
bobvh: No reason, added to gitignore by path for now, when the typescript part is moved into a dedicated gen-typescript I'll add a path with wildcards.

gen-python/src/python_api/jupyter_widget.rs
bobvh: Not sure what you mean there. It's a convenient way to have named colors, the ANSI basic colors (indexed), and absolute RGB). The python module itself doesn't force ratatui, the jupyter side is optional but requires ratatui.

bobvh: Moved the import to the top of the page regardless.

.gitignore
bobvh: done

bobvh:
Superseded by #131 which contains all of this work plus more.

PR #110, opened 2026-04-27
POC Creating new block groups in python without breaking node identity

bobvh:
This is something Chris at brought up a long time ago when we first had the networkx export: can we also go the other way around and create blocks directly in a networkx graph? We can with the Block datatype I implemented here for Python, and I used it to have an AI create a notebook that implements the "reverse translation graph" that we've been toying around with only in idea form.

(supersedes PR109 which had the wrong merge base, but it looks like github still lists all commits since main. Only commits from 5e9bae5 and onwards are relevant here)


dkhofer:
This demonstrates translating a protein sequence graph to the corresponding DNA sequence graph, right @bobvh ?

bobvh:
> This demonstrates translating a protein sequence graph to the corresponding DNA sequence graph, right @bobvh ?

Correct, it's demonstrated in the notebooks this PR added. Note that I pushed this as a POC, not intended to actually merge. Not deleting yet because I do want consider the upload-graph-from-networkx more if it fits in the API; or alternatively try to see if it's even possible to do the recodes without it.

PR #111, opened 2026-04-27
Better model error handling

dkhofer:
The goal here was to eliminate panic statements from model creation methods, motivated by me wanting to have a reasonable pattern to follow whenever creating a new model.


gen-models/src/accession.rs
```diff
@@ -255,51 +276,68 @@ impl Accession {
...
-            parent_accession_id: parent_accession_id.copied(),
-        })
+        match stmt.execute((hash, name, path_id, parent_accession_id)) {
+            Ok(_) => Ok(Accession {
+                id: hash,
+                name: name.to_string(),
+                path_id: *path_id,
+                parent_accession_id: parent_accession_id.copied(),
+            }),
+            Err(rusqlite::Error::SqliteFailure(err, _details))
```
dkhofer, line 300: The pattern I settled on for duplicates is to explicitly return an error in create methods, then handle that error in get_or_create by returning the object

gen-annotations/src/gff.rs
```diff
@@ -161,7 +161,8 @@ mod tests {
                 parent_block_group_id: None,
                 is_default: false,
             },
-        );
+        )
+        .unwrap();
```
dkhofer, line 167: Just a ton of new unwrap() calls in test methods

gen-models/src/accession.rs
```diff
@@ -20,6 +21,14 @@ pub struct Accession {
     pub parent_accession_id: Option<HashId>,
 }
 
+#[derive(Debug, Error, PartialEq)]
+pub enum AccessionError {
```
dkhofer, line 25: I'm on the fence about having these error definitions in the corresponding model class or in errors.rs

gen-models/src/block_group.rs
```diff
@@ -357,26 +391,24 @@ impl BlockGroup {
...
-                        parent_block_group_id: Some(&parent_block_group.id),
-                        ..Default::default()
-                    },
-                );
-                new_block_group.copy_contents_from(conn, parent_block_group);
-                new_block_group
-            })
-            .collect::<Vec<_>>();
+        let mut new_block_groups = vec![];
+        for parent_block_group in &parent_block_groups {
```
dkhofer, line 398: It's hard to get rust to compile when using the ? construct inside an iterator, so I just changed this to a for loop

src/updates/vcf.rs
```diff
@@ -536,17 +551,20 @@ pub fn update_with_vcf(
...
             let ref_start = vcf_entry.ref_start;
             let sequence =
-                SequenceCache::lookup(&mut sequence_cache, "DNA", vcf_entry.alt_seq.to_string());
+                SequenceCache::lookup(&mut sequence_cache, "DNA", vcf_entry.alt_seq.to_string())?;
             let sequence_string = sequence.get_sequence(None, None);
 
             let source_path_id = node_source_paths
                 .entry(vcf_entry.path.id)
                 .or_insert_with(|| {
-                    let block_group = BlockGroup::get_by_id(conn, &vcf_entry.block_group_id);
```
dkhofer, line 545: I changed BlockGroup::get_by_id to return a result, and that made it hard to rework this block of code to handle the error, so I decided to just cache all the data since it's probably best to avoid a database lookup for each row here anyway

src/lib.rs
```diff
@@ -90,6 +90,11 @@ pub fn track_database(
                 if let Some(uuid_db) = exist_by_uuid {
                     if path_db == uuid_db {
                         return Ok(());
+                    } else {
```
dkhofer, line 93: This fixes a bug I hit that caused a unit test to fail after I redid GenDatabase::create() to return a Result.  The unit test was looking for the error provided by this new else clause.  Before that change, the logic fell through to a place where it tried to create a GenDatabase, and that failed and returned an Err, which caused the unit test to pass, but for the wrong reason.

dkhofer:
This doesn't get rid of 100% of the panics in the models directory, but it takes care of most of them, and the rest should be relatively easy to clean up

gen-models/src/accession.rs
Chris7: I think we should put them here and consolidate/expose in errors.rs. So people can import error types easily.

gen-models/src/accession.rs
Chris7: yeah i like that as well

gen-models/src/accession.rs
```diff
@@ -255,51 +276,68 @@ impl Accession {
...
-                    let existing_id: HashId;
-                    if let Some(id) = parent_accession_id {
-                        existing_id = conn.query_row("select id from accessions where name = ?1 and path_id = ?2 and parent_accession_id = ?3;", params![name.to_string(), path_id, id], |row| row.get(0)).unwrap();
-                    } else {
-                        existing_id = conn.query_row("select id from accessions where name = ?1 and path_id = ?2 and parent_accession_id is null;", params![name.to_string(), path_id], |row| row.get(0)).unwrap();
-                    }
-                    Accession {
-                        id: existing_id,
+            Ok(accession) => Ok(accession),
+            Err(AccessionError::Duplicate(_)) => {
```
Chris7, line 319: Can we move the hash creation function in `create` to a function that can be used here so we can get the duplicate w/o going through the additional query? That woudl let us return the Accession directly. It looks like `id_hash` is the function name a few other places has used.

gen-models/src/block_group.rs
```diff
@@ -287,7 +318,8 @@ impl BlockGroup {
                 .into_iter()
                 .map(|edge| edge.id)
                 .collect::<Vec<_>>();
-            let new_path = Path::create(conn, &path.name, target_block_group_id, &edge_ids);
+            let new_path =
+                Path::create(conn, &path.name, target_block_group_id, &edge_ids).unwrap();
```
Chris7: meant to be a `?`?

gen-models/src/block_group.rs
Chris7: 👍

gen-models/src/files.rs
```diff
@@ -104,26 +116,20 @@ impl GenDatabase {
...
-                        "select * from gen_databases where db_uuid = ?1 AND name = ?2 AND path = ?3",
-                        params![db_uuid, name, path],
-                    ) {
-                        Ok(result) => Ok(result),
-                        Err(e) => Err(e),
-                    }
-                } else {
-                    panic!("something bad happened querying the database")
+            Err(GenDatabaseError::Duplicate(_)) => {
+                match GenDatabase::get(
```
Chris7, line 123: should return the constructed GenDatabase here

src/updates/fasta.rs
```diff
@@ -159,9 +160,9 @@ pub fn update_with_fasta(
                     state.first_node = None;
                 }
             }
-
-            change_count += 1;
         }
+
```
Chris7: was moving the variable here intentional?

src/updates/vcf.rs
Chris7: We could do it this way:
```
let source_path_id = if let Some(source_path_id) =
    node_source_paths.get(&vcf_entry.path.id)
{
    *source_path_id
} else {
    let block_group =
        BlockGroup::get_by_id(conn, &vcf_entry.block_group_id)?;
   ...
```

there's an extra lookup on the hash that way but it would let us get the ? using either approach. Or we could explode it to just the match arms

src/views/block_group.rs
```diff
@@ -245,7 +245,13 @@ pub fn view_block_group(
         std::collections::HashMap::new();
     let mut annotation_group_tracks: std::collections::HashMap<String, AnnotationTrack> =
         std::collections::HashMap::new();
-    let mut current_block_group = block_group_id.map(|bg_id| BlockGroup::get_by_id(conn, &bg_id));
+    let mut current_block_group =
+        block_group_id.map(|bg_id| match BlockGroup::get_by_id(conn, &bg_id) {
+            Ok(bg) => bg,
+            Err(err) => {
+                panic!("Failed to load block group {bg_id}: {err}");
```
Chris7, line 253: we should todo these so we handle them in the viewer gracefully

Chris7 (approved):
Looks great. Few comments on style. One place i think may be a bad move of a counter.

gen-models/src/accession.rs
dkhofer: Done

src/updates/vcf.rs
dkhofer: Good idea, done

gen-models/src/accession.rs
dkhofer: Good idea, done

gen-models/src/block_group.rs
dkhofer: Oops, yep, fixed

gen-models/src/files.rs
dkhofer: This is tricky, because while the UUID is unique, there isn't a uniqueness constraint on the other fields.  There's a test that specifically wants to get back whatever is associated with the uuid, even if the user tried to get-and-create with same UUID but different fields: https://github.com/genhub-bio/gen/blob/main/gen-models/src/files.rs#L252

I'm going to leave this as is but can update it in a followup PR if you have a preference.

src/updates/fasta.rs
dkhofer: Nope, fixed!

src/views/block_group.rs
dkhofer: Done

PR #112, opened 2026-04-28
Add toolchain

Chris7:
Define rust-toolchain so we're all on the same version


dkhofer (approved):
LGTM, let's see how it goes

PR #116, opened 2026-04-29
Add-file

Chris7:
Adds an add-file command for adding random files for pushing/pulling.

### DIFF
--- a/gen-models/src/operations.rs
+++ b/gen-models/src/operations.rs
@@ -1,14 +1,17 @@
 use std::{
     collections::HashMap,
     convert::TryInto,
+    fs,
     io::{self, BufReader},
     path::{Path, PathBuf},
     rc::Rc,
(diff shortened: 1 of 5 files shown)


gen-models/src/operations.rs
```diff
@@ -367,6 +372,123 @@ pub struct OperationInfo {
...
+        workspace,
+        &operation,
+        DatabaseChangeset {
+            db_path: gen_db.path,
+            changes: ChangesetModels::default(),
+        },
+        &DependencyModels::default(),
+    );
+
+    let gen_dir = workspace
```
Chris7, line 449: this stuff all gets removed by #119 so ignore it

dkhofer: Yep, :+1:

dkhofer (approved):
LGTM

PR #118, opened 2026-04-29
Add sequence error handler

Chris7:
Adds an error condition to fetching out of bound sequences. I added in support for circular sequence fetching, but didn't do a better implementation of storing the circles. For a later PR when we decide on sequence_type vs. a dedicated flag + some other stuff like rotating/offsets for tracking origins.

There are cascading updates from making result handlers. I've put comments in the areas with non-cosmetic changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 python: venv
 	.venv/bin/maturin develop --release --manifest-path gen-python/Cargo.toml --features extension-module --extras jupyter
 r:
-	docker build -t gen-r -f ./gen-r/Dockerfile .
+	docker build -q -t gen-r -f ./gen-r/Dockerfile .
 	docker run --rm gen-r
 clean:
 	cargo clean
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 python: venv
 	.venv/bin/maturin develop --release --manifest-path gen-python/Cargo.toml --features extension-module --extras jupyter
 r:
-	docker build -q -t gen-r -f ./gen-r/Dockerfile .
+	# Makes the build quiet for CI/agents
+	docker build -t gen-r -f ./gen-r/Dockerfile . 2>&1 | tail -n 25
 	docker run --rm gen-r
 clean:
 	cargo clean
--- a/gen-core/src/lib.rs
+++ b/gen-core/src/lib.rs
@@ -1,6 +1,5 @@
 use std::{convert::TryFrom, fmt, hash::Hash};
 
-use hex::FromHex;
 use rand::Rng;
 use sha2::{Digest, Sha256};
 
@@ -15,6 +14,7 @@ pub mod strand;
 pub mod traits;
 
 pub use config::Workspace;
+use errors::HashError;
 pub use generated::gen_core_capnp;
 pub use path::PathBlock;
 #[cfg(feature = "python-bindings")]
@@ -103,11 +103,10 @@ impl fmt::Display for HashId {
 }
 
 impl TryFrom<String> for HashId {
-    type Error = hex::FromHexError;
+    type Error = HashError;
 
     fn try_from(s: String) -> Result<Self, Self::Error> {
-        let bytes = <[u8; 32]>::from_hex(&s)?;
-        Ok(HashId(bytes))
+        HashId::try_from(s.as_str())
     }
 }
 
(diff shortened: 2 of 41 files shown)


gen-core/src/lib.rs
```diff
@@ -135,8 +135,9 @@ impl TryFrom<&str> for HashId {
     type Error = &'static str;
 
     fn try_from(s: &str) -> Result<Self, Self::Error> {
```
Chris7, line 136: There are some changes that were needed for the R bindings to pass tests again. Showed parts we missed.

gen-models/src/edge.rs
```diff
@@ -136,7 +136,7 @@ pub struct GroupBlock {
     pub id: i64,
     pub node_id: HashId,
     sequence: Option<String>,
-    external_sequence: Option<(String, String)>,
+    external_sequence: Option<(String, String, bool)>,
```
Chris7, line 139: this bool is tracking if the sequence is circular.

src/updates/gfa.rs
```diff
@@ -274,18 +274,30 @@ fn create_new_path_from_existing(
     let mut new_path_edges = vec![];
     let mut healing_edges = vec![];
     for (i, segment_id) in unmatched_path_segment_ids.iter().enumerate() {
-        if let Some((start, end)) = existing_path_ranges_by_segment_id.get(segment_id) {
+        if let Some((path_start, path_end)) = existing_path_ranges_by_segment_id.get(segment_id) {
```
Chris7, line 277: this is a legit bug in the code. start/end are actually positions in the entire path of the segment. But the code below was treating them as node sequence positions. I updated the var name to make this clear and updated the source/target coordinates to be node specific and not path.

Makefile
```diff
@@ -1,7 +1,7 @@
 python: venv
 	.venv/bin/maturin develop --release --manifest-path gen-python/Cargo.toml --features extension-module --extras jupyter
 r:
-	docker build -t gen-r -f ./gen-r/Dockerfile .
+	docker build -q -t gen-r -f ./gen-r/Dockerfile .
```
Chris7: quiets the docker build so it doesn't burn as many tokens

gen-core/src/lib.rs
dkhofer: :+1:

gen-models/src/edge.rs
dkhofer: Maybe it's time to convert it from a 3-tuple to a struct?

gen-models/src/sequence.rs
```diff
@@ -572,7 +740,7 @@ mod tests {
         for _ in 1..1_000_000 {
             let start = rand::rng().random_range(1..200_000_000);
 
-            sequence.get_sequence(start, start + 20);
+            sequence.get_sequence(start, start + 20).unwrap();
         }
         let elapsed = s.elapsed().as_secs();
         assert!(
```
dkhofer, line 746: Might be good to add tests confirming things work as expected with all the new error types

gen-tui/src/graph_controller.rs
```diff
@@ -1122,7 +1122,7 @@ mod tests {
         assert!(state.has_focus); // Focus is enabled by default for keyboard input
 
         state.focus();
-        state.camera_anim.is_some();
+        let _ = state.camera_anim.is_some();
```
dkhofer, line 1125: Thanks for fixing this!  It was bugging me too

src/updates/gaf.rs
```diff
@@ -116,15 +116,22 @@ where
...
+        if let Some(node_info) = node_lengths.get(node_id) {
+            return Some(*node_info);
+        }
+
+        let node_info: Vec<&str> = node_id.rsplitn(2, '.').collect();
+        let node_id = *node_info.last().unwrap();
+        let id = HashId::try_from(node_id).ok()?;
+        let mut stmt = conn
+            .prepare_cached(
+                "select s.length from nodes n left join sequences s on (s.hash = n.sequence_hash) where n.id = ?1;",
```
dkhofer: I know it's a bit of a tangent, but I feel like this part of the code belongs in the Node model instead of the update with gaf code

src/updates/gfa.rs
dkhofer: :+1:

dkhofer (approved):
A few minor comments, but LGTM

Makefile
dkhofer: :+1:

gen-models/src/edge.rs
Chris7: Absolutely. I want to do that in a followup when we agree on how to do circles. This just put in the basic mechanisms for circular indexing.

src/updates/gaf.rs
Chris7: i can move it. I wanted to reduce the new stuff but if you're alright w/ it

PR #119, opened 2026-04-30
Consoldiate file addition

Chris7:
We were repeating this in a few places, I put it under FileAddition

### DIFF
--- a/gen-models/src/operations.rs
+++ b/gen-models/src/operations.rs
@@ -1,14 +1,17 @@
 use std::{
     collections::HashMap,
     convert::TryInto,
+    fs,
     io::{self, BufReader},
     path::{Path, PathBuf},
     rc::Rc,
     string::ToString,
 };
 
+use anyhow::anyhow;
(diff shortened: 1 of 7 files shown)


gen-models/src/operations.rs
```diff
@@ -2829,7 +2836,7 @@ mod tests {
 
         let db_uuid = metadata::get_db_uuid(graph_conn);
         GenDatabase::create(operation_conn, &db_uuid, "default", "default.db").unwrap();
-        Branch::get_or_create(operation_conn, "main");
+        let _ = Branch::get_or_create(operation_conn, "main");
```
dkhofer: It'd be better to call .unwrap() here, then you don't have to use `let _ =` and it'll cause the test to fail with an error

dkhofer: (If the result is an error)

dkhofer (approved):
One minor comment, otherwise LGTM!

PR #122, opened 2026-05-01
Update filepath for out-of-directory files and track filename

Chris7:
If a file outside of the gen directory is added, we store the absolute path to it. Given the asset store changes, we can instead refer to the copy of the file in assets. Given these assets have checksums as names, I added a new column to OperationFiles to track the filename.

### DIFF
--- a/gen-core/src/config.rs
+++ b/gen-core/src/config.rs
@@ -6,6 +6,7 @@ use std::{
 use crate::{HashId, errors::ConfigError};
 
 pub const CHANGESET_DIR_NAME: &str = "changesets";
+pub const ASSETS_DIR_NAME: &str = "assets";
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Workspace {
@@ -32,6 +33,8 @@ impl Workspace {
         ensure_dir(&gen_path);
         let changesets = gen_path.join(CHANGESET_DIR_NAME);
         ensure_dir(&changesets);
+        let assets = gen_path.join(ASSETS_DIR_NAME);
+        ensure_dir(&assets);
         gen_path
     }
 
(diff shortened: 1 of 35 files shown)


gen-core/src/config.rs
```diff
@@ -62,6 +65,13 @@ impl Workspace {
             .ok_or(ConfigError::GenDirectoryNotFound)
     }
 
+    pub fn asset_dir(&self) -> Result<PathBuf, ConfigError> {
```
Chris7, line 68: there's a few places all getting the assets dir. It lives here now

gen-models/src/operations.rs
```diff
@@ -500,38 +523,128 @@ impl FileAddition {
         HashId(calculate_hash(&combined))
     }
 
-    fn normalize_file_paths(workspace: &Workspace, file_path: &str) -> (String, String) {
```
Chris7, line 503: i broke this into 2 functions. It was getting called for just getting one of the two in places and that seemed bad.

gen-models/src/operations.rs
```diff
@@ -500,38 +523,128 @@ impl FileAddition {
...
+    fn asset_filename(checksum: &HashId, file_type: FileTypes) -> String {
```
Chris7, line 532: more normalization and consolidation of where asset names/paths are coming from

gen-core/src/config.rs
dkhofer: :+1:

gen-models/src/operations.rs
dkhofer: :+1:

gen-models/src/operations.rs
dkhofer: :+1:

dkhofer (approved):
All seems reasonable, LGTM

dkhofer:
Although, why are those two checks failing?

PR #123, opened 2026-05-04
fix(gen-python): replace db_path with DbContext

bobvh:
DbHandle::path() was removed upstream, breaking PyRepository which stored a PathBuf to re-open connections and stamped that path onto every PyBlockGroup so they could open their own connections for plot().

  Replace both with DbContext:
  - PyRepository stores DbContext directly (eager open, no lazy Mutex init)
  - PyBlockGroup carries `Option<DbContext>` instead of `Option<PathBuf>`
  - Remove with_connection() closure pattern, now dead
  - Rename bg() -> into_py_block_group() to clarify intent
  - Drop stale DbHandle::path() tests; add Python-facing read method tests

(+ fix upstream clippy warning)


Chris7 (approved):
lgtm

PR #124, opened 2026-05-04
feat(gen-tui): cell-level highlight support

bobvh:
## Human tl;dr
We already had 3 ways to highlight specific subsets of a graph: by node, by edge and by path
This PR adds "by cells" within a node (so by character) and adds a function to move the camera to exactly that spot.

## Claude
Adds a Cells variant to HighlightKind and a cell_highlights Vec to CroppedGraph, enabling sub-rectangle tinting of individual graph nodes. The tint pass runs after normal node rendering and blends onto whatever characters are already in the WorldBuffer.

Adds two public helpers to gen_graph_widget:
>   center_on_node_offset — loads the owning partition, sets it as anchor,
>   resolves the domain NodeIndex, and snaps the camera directly to the
>   node's world position at a fractional (x, y) offset.

>   highlight_match_range — iterates a GraphLocus, calls set_cell_highlight
>   on the start/end nodes with the correct column slice, fully tints middle
>   nodes, and adds edge highlights between consecutive locus blocks.

Together these let a caller navigate to a sequence search hit and paint exactly the matched columns — start offset on the first node, full width on interior nodes, end offset on the last — in a single call.

### DIFF
--- a/gen-tui/src/graph_controller.rs
+++ b/gen-tui/src/graph_controller.rs
@@ -337,8 +337,9 @@ where
         color
     }
 
-    /// Internal helper to apply a node rect highlight to the viewport graph.
-    /// Stores the domain NodeIndex directly — no world-coordinate conversion needed.
+    /// Render-only primitive shared by `set_cell_highlight` (initial paint) and the
+    /// highlight replay loop (re-paint after viewport rebuild). Kept separate so replay
+    /// can call it without pushing duplicate entries into `self.highlights`.
     fn apply_cell_highlight(
         viewport_graph: &mut CroppedGraph,
         graph: &G,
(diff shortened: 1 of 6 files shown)


gen-tui/src/graph_controller.rs
```diff
@@ -79,6 +79,13 @@ pub enum HighlightKind<N> {
     Edge(N, N),
     /// A path consisting of a sequence of nodes
     Path(Vec<N>),
+    /// Tint a sub-rectangle of a single node.
+    /// tl/br are (col, row) offsets from the node's top-left corner; br is exclusive.
+    Cells {
+        node: N,
+        tl: (i64, i64),
```
Chris7, line 87: please add offset to this name, tl means do this at top left, same with bottom right

Chris7, line 87: don't call this an offset in the comment. It's a position, i don't think of "x" as "x offset from y axis"

Chris7 (approved):
seems minor, just update the comment wrt offsets

gen-tui/src/graph_controller.rs
bobvh: My reasoning was that they're relative coordinates, but you're right that for the user of the function that doesn't matter. I'll reword them so it's clear that we're talking about the corners of the part to highlight.

PR #125, opened 2026-05-05
Annotation parents

Chris7:
Looks at ancestors and shows annotation groups present on them

### DIFF
--- a/src/imports/genbank.rs
+++ b/src/imports/genbank.rs
@@ -738,7 +738,12 @@ mod tests {
 
     use super::*;
     use crate::{
-        test_helpers::setup_gen, track_database, views::annotations::load_annotations_for_group,
+        test_helpers::setup_gen,
+        track_database,
+        views::{
+            annotation_groups::{AnnotationGroupEntry, AnnotationGroupOrigin},
+            annotations::{AnnotationGroupTrackRequest, load_annotations_for_group},
+        },
     };
 
     fn get_unmodified_sequence() -> String {
@@ -1190,8 +1195,20 @@ mod tests {
                 .push((block.sequence_start, block.sequence_end));
         }
 
-        let spans =
-            load_annotations_for_group(conn, &groups[0].name, &visible_ranges_by_node).unwrap();
+        let entry = AnnotationGroupEntry {
+            id: format!("{}::{}", block_group.sample_name, groups[0].name),
+            name: groups[0].name.clone(),
+            sample_name: block_group.sample_name.clone(),
+            source_block_group_id: block_group.id,
+            origin: AnnotationGroupOrigin::CurrentSample,
+        };
+        let spans = load_annotations_for_group(&AnnotationGroupTrackRequest {
+            conn,
+            current_block_group: &block_group,
+            entry: &entry,
+            visible_ranges_by_node: &visible_ranges_by_node,
+        })
+        .unwrap();
         let m13_forward = spans
             .iter()
             .find(|annotation| annotation.name == "M13 Forward")
--- a/src/imports/genbank.rs
+++ b/src/imports/genbank.rs
@@ -6,9 +6,10 @@ use std::{
 };
 
 use gb_io::reader;
+use gen_annotations::projection::AnnotationSegment;
 use gen_core::{
     HashId, PATH_END_NODE_ID, PATH_START_NODE_ID, PathBlock, Strand,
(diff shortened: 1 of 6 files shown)


src/views/block_group.rs
```diff
@@ -200,18 +201,14 @@ pub fn view_block_group(
         None
     };
 
-    // Create explorer and its state that persists across frames
```
Chris7, line 203: we store a reference to the block group in here now so this is just moved below

src/views/block_group.rs
```diff
@@ -1082,8 +1126,15 @@ pub fn view_block_group(
...
                         }
                         annotation_group_tracks.insert(
-                            entry.name.clone(),
-                            AnnotationTrack::new(entry.name.clone(), spans),
+                            entry.id.clone(),
+                            AnnotationTrack::new(
+                                if entry.sample_name == bg.sample_name {
+                                    entry.name.clone()
+                                } else {
+                                    format!("{} ({})", entry.name, entry.sample_name)
```
Chris7, line 1134: this makes the header line for the track have the ancestor sample name in () if loaded

src/views/annotation_groups.rs
```diff
@@ -1,16 +1,211 @@
...
     pub name: String,
+    pub sample_name: String,
+    pub source_block_group_id: HashId,
+    pub origin: AnnotationGroupOrigin,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum AnnotationGroupOrigin {
+    CurrentSample,
+    ParentSample,
```
dkhofer, line 21: Why does ParentSample merit a separate entry here?

src/views/block_group.rs
dkhofer: :+1:

src/views/block_group.rs
dkhofer: :+1:

src/imports/genbank.rs
```diff
@@ -738,7 +738,12 @@ mod tests {
 
     use super::*;
     use crate::{
-        test_helpers::setup_gen, track_database, views::annotations::load_annotations_for_group,
+        test_helpers::setup_gen,
+        track_database,
+        views::{
```
dkhofer: I kind of recoil at having an import operation depend on code under the views directory.  Could you refactor all the common stuff that's used here out of src/views and into the gen-annotations subcrate?  It looks like it'll be mostly but not 100% clean to do that, so might be a little extra work.  But I don't like having dependencies between two parts of the codebase that are orthogonal

dkhofer (changes requested):
I think there's some important refactoring needed before I feel comfortable approving

src/views/annotation_groups.rs
Chris7: in the viewer there's an immediate parent vs ancestor distinction in the layout

src/imports/genbank.rs
Chris7: Looking at it, a lot of what is in the view is specific to how we want to view something. I'm going to try and just not import these vs. move them atm.

src/imports/genbank.rs
```diff
@@ -76,26 +77,6 @@ struct FinalSequenceSegment {
     sequence_range: Range,
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
```
Chris7: i found these were redundant w/ gen-annotations too, so that has been leveraged here.

src/imports/genbank.rs
```diff
@@ -737,9 +710,7 @@ mod tests {
     use noodles::fasta;
 
     use super::*;
-    use crate::{
-        test_helpers::setup_gen, track_database, views::annotations::load_annotations_for_group,
```
Chris7: dropped the view stuff and now leverage gen-annotations

Chris7:
the build is failing in a weird way. It's failing on code that is deleted, i'm not sure there isn't something wonky w/ the runners cache

src/imports/genbank.rs
dkhofer: Cool!

src/imports/genbank.rs
dkhofer: Great, thanks

dkhofer (approved):
LGTM!

PR #126, opened 2026-05-06
Uri support

Chris7:


gen-models/src/assets.rs
```diff
@@ -0,0 +1,926 @@
...
+
+    fn io_path(&self) -> PathBuf {
+        self.source_path
+            .clone()
+            .unwrap_or_else(|| PathBuf::from(self.file_path()))
+    }
+
+    fn calculate_checksum(&self, path: &str) -> Result<HashId, FileAdditionError> {
+        let (op, op_path) = local_operator(Path::new(path)).map_err(opendal_file_addition_error)?;
+        if let Err(err) = op.stat(&op_path) {
```
dkhofer: It seems to me this stat isn't doing anything that is different from the next block that tries to create a file reader.  Each block has exactly the same pattern of error handling.  Is this block necessary?

gen-models/src/assets.rs
```diff
@@ -0,0 +1,926 @@
...
+        }
+
+        if provided_path.is_absolute() {
+            return Self::asset_relative_path(workspace, checksum, file_type);
+        }
+
+        Ok(file_path)
+    }
+
+    /// This exists along with store_file because one uses the model Self attributes to
```
dkhofer: The wording for this comment is a bit hard to parse.  It would be clearer to say store_file uses an existing FileAddition object to determine where to store the file, while this method uses the provided source_path

dkhofer (approved):
Overall LGTM, comments are minor

gen-models/src/assets.rs
Chris7: yeah, nuked it

PR #127, opened 2026-05-06
Clone command

dkhofer:


src/operation_management.rs
```diff
@@ -1040,6 +1041,8 @@ fn pull_from_remote_server(
         );
     }
 
+    println!("Pull succeeded from {remote_url}");
```
dkhofer, line 1091: I felt like cleaning this up.  The output from the pull command was a bit weird otherwise

dkhofer:
Taking this back to put in logic to pull and if there's an authorization failure, only attemp to log in then

src/commands/clone.rs
```diff
@@ -0,0 +1,190 @@
...
+    track_database,
+};
+
+const ORIGIN: &str = "origin";
+
+pub fn execute(url: &str) -> Result<(), Box<dyn std::error::Error>> {
+    let repo_name = infer_repo_name(url)?;
+    let repo_path = PathBuf::from(&repo_name);
+
+    fs::create_dir(&repo_path)?;
```
Chris7: do we want to error if the dir already exists?

src/commands/clone.rs
```diff
@@ -0,0 +1,190 @@
...
+};
+
+use crate::{
+    commands::remote::login_remote,
+    get_connection, get_operation_connection,
+    operation_management::{RemoteOperationError, pull},
+    track_database,
+};
+
+const ORIGIN: &str = "origin";
```
Chris7, line 20: Idea for this or a follow up:
* Move this to like src/lib.rs
* Update remote add to set this as the default name, so `gen remote add XXX` just works to do that
* Have the first remote added that way automatically be set to default

Chris7 (approved):
All looks good! One comment on behavior if a dir already exists

src/commands/clone.rs
dkhofer: I'm going to do that in a followup, and then update the help instructions on the blank repo page accordingly.  It doesn't make sense for it to be more than one or two lines to push a new repo

src/commands/clone.rs
dkhofer: Good idea, done

PR #130, opened 2026-05-07
feat(gen-tui): annotation overlay and track rendering

bobvh:
## Summary

This PR adds two distinct annotation systems to the graph viewer.

### Track-based annotations

Track-based annotations appear below the graph as a dedicated panel (technically an overlay, not a panel). We extended the packing algorithm to support labels and strand indicators alongside the existing interval rendering. The panel expands automatically as tracks are added, so callers don't need to manage layout dimensions manually.

### Inline annotations

Inline annotations are rendered directly on the graph nodes. This mode is primarily intended for use in Python notebooks (see follow-up PR), where it has proven useful for displaying search results and variant-specific annotations in libraries — a use case that is significantly harder to express cleanly in a track-based layout.

🤖 Generated with [Claude Code](https://claude.com/claude-code) (partially, he's really running with the credit here)

### DIFF
--- a/gen-tui/src/graph_controller.rs
+++ b/gen-tui/src/graph_controller.rs
@@ -294,6 +294,25 @@ where
         }
     }
 
+    /// Internal helper to apply a node rect highlight to the viewport graph.
+    /// Resolves the node to its world position and stores that for consistency
+    /// with node_highlights and edge_highlights.
+    fn apply_cell_highlight(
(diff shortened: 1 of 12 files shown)


bobvh:
Merging this, Chris already saw it. I may do some refactoring in the future but for now I want to get this PR through as is.

PR #131, opened 2026-05-07
feat(gen-python): Python bindings refactor, TypeScript widget, and TUI fixes

bobvh:
### Bindings refactor

Splits the monolithic `repository.rs` into a `repository/` subdirectory with focused modules: `mod.rs` (core block group queries, plot, get_block_sequence), `imports.rs`, `exports.rs`, `updates.rs`, `graph_ops.rs` (derive_chunks, subgraph, make_stitch), and `search.rs`. Top-level standalone functions (`imports.rs`, `exports.rs`, `updates.rs`, `graph_operations.rs`) are removed — all superseded by `Repository` methods.

`Factory` struct removed; `subgraph` and `chunks` moved directly onto `PyBlockGroup`. Adds `graph_search.rs` with `PyGraphLocus`, `PyGraphPos`, and `PyAnnotation` for sequence search and navigation.

**API cleanup:** `DbContext` replaces `db_path` throughout; `SequenceKind` replaces `case_sensitive` bool; `parse_sequence_kind` deduplicated across modules.

**PyO3 modernization:** public API names cleaned up via `#[pyclass(name = "...")]` (e.g. `PyBlockGroup` → `BlockGroup`); `__eq__` updated to `&Bound<'_, PyAny>`; `Python::with_gil` removed from `#[pymethods]`.

### Jupyter widget

TypeScript anywidget backed by a Rust serialization layer streaming rendered frames to the browser. Sources in `gen-python/js/` draw cells, box-drawing chars, and text onto a canvas with pan/drag, zoom, split-glyph/circle node rendering, inline annotation rendering, and track panels. Compiled JS checked in — npm not required at build time.

These commits are also present in PR #108 (jupyter-widget-ts), which has already been reviewed and approved. Reviewers can treat this section as pre-approved.

### TUI fixes

Fixes to `graph_controller`, `graph_widget`, `plotter`, `viewport_graph`, and the `src/views/` layer to support the annotation overlay model and widget cursor/freeze behavior.

---

**Stacking note:** depends on PR #130 (tui-annotations); targets that branch directly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gen-python/src/python_api/repository/mod.rs
```diff
@@ -22,41 +22,55 @@ pub mod imports;
 pub mod search;
 pub mod updates;
 
+fn tx_begin(context: &DbContext) -> PyResult<()> {
```
bobvh, line 25: All the transaction management for the Python bindings is currently consolidated here. They're called by run_write() and the repo.transaction() context manager. 

- run_write is called by every individual write method (import_fasta, update_with_vcf, etc.). it starts a new transaction unless the in_transaction flag indicates there is already one ongoing.

- with repo.transaction() starts a Python context manager that groups multiple methods in the same DB transaction, so that if one of them raises an exception, the whole things rolls back.

gen-python/src/python_api/repository/mod.rs
```diff
@@ -22,41 +22,55 @@ pub mod imports;
...
+        .map_err(sqlite_err_to_pyerr)?;
+    Ok(())
+}
+
+fn tx_rollback(context: &DbContext) {
+    context.graph().conn().execute("ROLLBACK", []).ok();
+    context.operations().conn().execute("ROLLBACK", []).ok();
+}
+
 pub(crate) fn run_write<F, T>(context: &DbContext, managed: bool, op: F) -> PyResult<T>
```
bobvh, line 57: Called by all methods that result in db writes so they can start a new transaction or continue the one that's ongoing.

gen-python/src/python_api/repository/mod.rs
```diff
@@ -155,17 +169,7 @@ impl PyRepository {
     }
 
     fn __enter__(mut slf: PyRefMut<'_, Self>) -> PyResult<()> {
```
bobvh, line 171: This is where the `with(repo.transaction ())`: context manager is set up.

gen-python/src/python_api/jupyter_widget.rs
```diff
@@ -67,76 +93,186 @@ fn indexed_to_hex(i: u8) -> String {
...
 }
 
 /// Internal graph controller for the Jupyter notebook widget.
 ///
 /// Not intended for direct use from Python — users should call
 /// `repo.plot(bg)` or `bg.plot()` which return a `GenGraphWidget`.
+///
+/// # Thread safety
+///
+/// ipykernel 6+ runs cell code in a thread-pool executor (e.g. thread 12) while
```
bobvh, line 214: I didn't love having to do this (initialize with a path to the db file) after spending all that time to use DbContext everywhere. What's the current multi-threading plan for the rest of the codebase?

gen-python/src/python_api/repository/mod.rs
bobvh: @dkhofer the managed flag is how the "am I in a context manager or not" is handled.

gen-python/src/python_api/repository/mod.rs
dkhofer: This is all good.  Once this is in I'd like to take a stab at pushing run_write etc. down into code somewhere under the src/ directory, to be used directly in things like src/import/fasta.rs.  That's one thing I was getting at when we talked today.  I don't think we should reimplement it in shell/python/R/whatever other bindings we provide

gen-python/src/python_api/repository/mod.rs
dkhofer: :+1:

gen-python/src/python_api/repository/mod.rs
dkhofer: :+1:

gen-python/src/python_api/jupyter_widget.rs
dkhofer: I don't think we have one, because from the standpoint of only calling gen on the command line, everything is single threaded.

That said, I think you could use a thread-local cache to load the DbContext once, store it in the cache, and then access it (probably using the db path as the key).  Maybe ask codex to try it out?  This seems to be the cache to use. https://docs.rs/local_lru/latest/local_lru/

dkhofer (approved):
All looks reasonable.  The pattern you're using to handle transactions is pretty common and the way it's implemented is pretty standard.  LGTM!

PR #133, opened 2026-05-11
New region parser

Chris7:
Update region parser to:
* accept negative coordinates
* wrap around support where start > end
* use it in some places where we aren't

I used the new parser in a handful of places. It was confusing having the region define start/end and also be able to pass it as a flag so the flags were removed. Current setup doesn't change the input types allowed for functions (e.g. paths still needed). Next change would be to implement support for making changes from other types and leverage the insert_changes update.


gen-core/src/region.rs
```diff
@@ -27,26 +29,79 @@ pub enum RegionParseError {
...
+            Some(coordinates) => {
+                if coordinates.is_empty() {
+                    return Err(RegionParseError::InvalidSyntax);
+                }
+
+                let bytes = coordinates.as_bytes();
+                let separator = bytes
+                    .iter()
+                    .enumerate()
+                    .skip(1)
```
Chris7: this handles the first `-` and we do this general approach so we can have a span to a negative, like `-2--3`

src/commands/update/library.rs
```diff
@@ -25,10 +25,10 @@ pub struct Command {
     path_name: String,
     /// The start coordinate for the region to add the library to
     #[arg(long)]
-    start: i64,
+    start: Option<i64>,
```
Chris7: this is redundant w/ the region parser, remove it?

src/commands/update/fasta.rs
```diff
@@ -25,12 +25,6 @@ pub struct Command {
     /// The name of the region to update (eg "chr1")
     #[arg(long)]
     region_name: String,
-    /// The start coordinate for the region to add the library to
```
Chris7, line 28: these are redundant now with the region parser and confusing to keep in

gen-annotations/src/lib.rs
```diff
@@ -1,5 +1,4 @@
 pub mod gff;
-pub mod projection;
```
Chris7, line 2: projection was moved to gen-models. This was done because of circular imports.

gen-core/src/region.rs
```diff
@@ -2,11 +2,11 @@ use std::num::ParseIntError;
 
 use thiserror::Error;
 
-#[derive(Debug, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
```
Chris7, line 5: changes are:
* making start/end optional
* supporting wrap around
* negative indexing

gen-core/src/region.rs
```diff
@@ -25,28 +27,97 @@ pub enum RegionParseError {
...
+pub enum RegionResolutionError<E: std::error::Error + 'static> {
+    #[error("Region not found: {0}")]
+    NotFound(String),
+    #[error("Region is ambiguous: {0}")]
+    Ambiguous(String),
+    #[error(transparent)]
+    Lookup(#[from] E),
+}
+
+pub trait RegionResolver: Sized {
```
Chris7, line 40: this is the model trait to implement to resolve a region

gen-models/src/block_group.rs
```diff
@@ -192,7 +194,7 @@ impl IntervalTreeSource for Path {
         &self,
         conn: &GraphConnection,
     ) -> Result<IntervalTree<i64, NodeIntervalBlock>, BlockGroupError> {
-        self.intervaltree(conn).map_err(Into::into)
+        Path::intervaltree(self, conn).map_err(Into::into)
```
Chris7, line 197: cargo check didn't like the self vs. Path and similar in other models

gen-models/src/region.rs
```diff
@@ -0,0 +1,349 @@
...
+    anchor_end: i64,
+}
+
+pub fn resolve(
+    region: &Region,
+    conn: &GraphConnection,
+    collection_name: &str,
+    sample_name: &str,
+) -> Result<ResolvedGenRegion, GenRegionError> {
+    match resolve_block_group(region, conn, collection_name, sample_name) {
```
Chris7, line 90: these are all just done in order of probably importance. If something wanted to really only look for annotations/etc. they wouldn't use this helper

gen-models/src/region.rs
```diff
@@ -0,0 +1,349 @@
...
+    match resolve_annotation(region, conn, collection_name, sample_name) {
+        Ok(region) => return Ok(region),
+        Err(GenRegionError::NotFound(_)) => {}
+        Err(err) => return Err(err),
+    }
+
+    resolve_accession(region, conn, collection_name, sample_name)
+}
+
+pub fn resolve_path(
```
Chris7, line 111: these additional methods are here instead of in the models because they serve back the blockgroup/etc. Here the locations are figured out

gen-models/src/region.rs
```diff
@@ -0,0 +1,551 @@
...
+        },
+        anchor_start: target.anchor_start,
+        anchor_end: target.anchor_end,
+        feature_length: target.feature_length,
+        start,
+        end,
+    })
+}
+
+fn target_from_accession(
```
Chris7, line 282: this exists because multiple things wrap accessions (accession and annotation atm, path in the future probably)

gen-models/src/region.rs
```diff
@@ -0,0 +1,551 @@
...
+}
+
+impl ResolvedGenRegion {
+    pub fn offset_range(
+        &self,
+        start_offset: i64,
+        end_offset: i64,
+    ) -> Result<(i64, i64), GenRegionError> {
+        let (start, end) = match self.kind {
+            ResolvedRegionKind::Annotation | ResolvedRegionKind::Accession => (
```
Chris7, line 313: annotations will be relative to themselves in the graph, so we do the offset

gen-python/src/python_api/repository/updates.rs
```diff
@@ -19,16 +19,14 @@ use crate::python_api::sequence_part::PySequencePart;
 
 #[pymethods]
 impl PyRepository {
-    #[pyo3(signature = (filename, sample, new_sample, region_name, start, end, name=None))]
```
Chris7, line 22: bunch of signature updates for removing start/end in python and r

src/views/annotations.rs
```diff
@@ -27,60 +30,15 @@ use crate::views::{
 };
 
 fn accession_edges_to_segments(edges: &[AccessionEdge]) -> Vec<AnnotationSegment> {
```
Chris7, line 32: i want to do another change where we use AnnotationSegment from gen-annotation instead of the view so this can be removed. tbd

gen-core/src/region.rs
dkhofer: :+1:

gen-core/src/region.rs
dkhofer: :+1:

gen-core/src/region.rs
```diff
@@ -25,28 +27,121 @@ pub enum RegionParseError {
...
+                    };
+                    let end = if end.is_empty() {
+                        None
+                    } else {
+                        Some(
+                            end.parse::<i64>()
+                                .map_err(RegionParseError::InvalidCoordinate)?,
+                        )
+                    };
+                    if start.is_none() && end.is_none() {
```
dkhofer, line 108: Seems like the if and else clauses are the same thing?

gen-core/src/region.rs
```diff
@@ -25,28 +27,121 @@ pub enum RegionParseError {
...
+            None => (region_string.trim(), None),
+        };
+
+        if name.is_empty() {
+            return Err(RegionParseError::NoName);
         }
 
-        Ok(Self { name, start, end })
+        let (start, end) = match coordinates {
+            Some(coordinates) => {
```
dkhofer, line 84: Could you add a high-level comment describing the parsing strategy here?  Seems like you're looking for ".." and then if that's not present, looking for "-", and splitting on the found separator?

gen-core/src/region.rs
```diff
@@ -25,28 +27,121 @@ pub enum RegionParseError {
...
+                                .map_err(RegionParseError::InvalidCoordinate)?,
+                        )
+                    };
+                    if start.is_none() && end.is_none() {
+                        (None, None)
+                    } else {
+                        (start, end)
+                    }
+                } else {
+                    let bytes = coordinates.as_bytes();
```
dkhofer, line 117: Is this doing a different thing from the code for splitting on ".."?  If not, what is different?  If it's the same, I prefer what's used for ".."

gen-models/src/block_group.rs
dkhofer: :+1:

gen-models/src/region.rs
dkhofer: :+1:

gen-models/src/region.rs
dkhofer: :+1:

gen-models/src/region.rs
dkhofer: :+1:

gen-models/src/region.rs
dkhofer: :+1:

gen-python/src/python_api/repository/updates.rs
dkhofer: :+1:

dkhofer: I think it'd be good to rename `region_name` to `region` for all these, too

src/commands/update/fasta.rs
dkhofer: :+1:

src/graphs/operators.rs
```diff
@@ -280,17 +287,11 @@ fn get_block_group_id(
...
-    for block_group in &block_groups {
-        if block_group.name == region_name {
-            return Ok(block_group.id);
-        }
-    }
-
-    Err(GraphOperationError::RegionNotFound(format!(
-        "No region found with name: {region_name}"
-    )))
+    let resolved_region = Region::parse(region_name)
```
dkhofer, line 290: Glad to see this cleaned up!

src/views/annotations.rs
dkhofer: :+1:

dkhofer (approved):
Minor comments, but LGTM

gen-core/src/region.rs
Chris7: they won't be, the first if is for both None, the else is mixed

gen-core/src/region.rs
Chris7: it's a bit different because of the negative sign is a dash which is also the separator

gen-core/src/region.rs
Chris7: yeah, added that info. I think we can rewrite/adjust it. The useful thing is having the region parser struct used in more places

PR #134, opened 2026-05-11
Fix mac os file bugs

dkhofer:


gen-python/package.json
```diff
@@ -6,7 +6,7 @@
     "check": "tsc --noEmit"
   },
   "devDependencies": {
-    "esbuild": "^0.24.0",
+    "esbuild": "^0.25.0",
```
dkhofer, line 9: Fixes this: https://github.com/genhub-bio/gen/security/dependabot/45

gen-models/src/assets.rs
```diff
@@ -636,7 +636,15 @@ impl LocalAssetUri {
     }
 
     fn canonicalize_or_normalize(path: &Path) -> PathBuf {
-        fs::canonicalize(path).unwrap_or_else(|_| Self::normalize_path(path))
+        let normalized_path = fs::canonicalize(path).unwrap_or_else(|_| Self::normalize_path(path));
+        // Unfortunately, the path we get back from Mac OS X sometimes starts
```
bobvh, line 640: Is the "sometimes" specifically when you're getting a path from /tmp or also for assets from subdirectories of /Users?

bobvh: The way I understand it /private was supposed to be OS-level stuff, which happens to include tmp, so /tmp is a symlink to /private/tmp. As long as everything that uses the paths follows the symlinks your hack should be fine.

dkhofer: For me, it's paths with /var at the root, which of course is also an OS-level directory.

bobvh (approved):
It's a workaround, but good to go here

PR #135, opened 2026-05-12
fix(derive_chunks): off-by-one in chunk point generation + add tests

bobvh:
`(0..chunk_count)` produced 0 as the first breakpoint, causing a 0..0 range that yielded an empty interval-tree query and a panic in operators.rs. Changed to `(1..=chunk_count)` with guards and added 3 tests.

### DIFF
--- a/src/commands/graph_operations/derive_chunks.rs
+++ b/src/commands/graph_operations/derive_chunks.rs
@@ -87,8 +87,9 @@ pub fn derive_chunks_operation(
         result.into_iter().sorted().collect::<Vec<i64>>()
     } else if let Some(chunk_size) = chunk_size {
         let chunk_count = path_length / chunk_size;
-        (0..chunk_count)
+        (1..=chunk_count)
             .map(|i| i * chunk_size)
+            .filter(|&p| p < path_length)
             .collect::<Vec<i64>>()
     } else {
         return Err(DeriveChunksOperationError::NoChunkingMethod(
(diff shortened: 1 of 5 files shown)


src/commands/graph_operations/derive_chunks.rs
```diff
@@ -87,23 +87,17 @@ pub fn derive_chunks_operation(
...
+            .filter(|&p| p < path_length)
             .collect::<Vec<i64>>()
     } else {
         return Err(DeriveChunksOperationError::NoChunkingMethod(
             "No chunking method specified.".to_string(),
         )
         .into());
     };
-
-    if chunk_points.is_empty() {
```
dkhofer, line 100: I think this is still a legitimate error condition?  Why is it getting removed?

dkhofer (commented):
The bug fix and tests look good, just one concern about some logic that I think is unrelated and getting deleted

src/commands/graph_operations/derive_chunks.rs
bobvh: What's your expectation of the outcome of asking to break a piece of 100 bp into chunks of 200 bp? Should that be an error? I was thinking that should just return one chunk, but can be convinced otherwise.

That's the case where you en up here, right?

dkhofer: One way this could happen is if there's a legitimate bug somehow and the user managed to specify an empty list.  I think they'd want to get an error in that case.  I worry about that more with the other possible code path that could lead here, parsing the breakpoints list (if provided).

For your case of breaking a segment of 100 bp, that's actually also my concern, like if they don't actually realize they're asking for a chunk size larger than the contig, it may be helpful for them to get an error.  But if you think the more common case is for them to want gen to do what they mean, that could work.

If so, my preference would be for you to add an if statement in the block that sets up the chunk points given a chunk size to check if the size is greater than the segment, and if so create a non-empty chunk points list that is the range of the sequence being split up.

bobvh: Got it, it's between:
- better to have the likely python user try and catch the exception (or measuring beforehand) when they're breaking something up into chunks 
- if we still do it silently, make sure that the correct process is followed

dkhofer: Yeah, and also fwiw this affects the command line too.  I just wanted to make sure it wasn't an unintentional deletion.  Since it could go either way, I'll leave it to you to handle, and approve this PR

dkhofer (approved):
LGTM

PR #137, opened 2026-05-12
Remove path from accession

Chris7:
Drops paths for block_group for accessions. We don't even need block_groups to be honest but we can do that if it ever makes sense. This is so we can label parts outside of paths

### DIFF
--- a/gen-models/migrations/core/01-initial/up.sql
+++ b/gen-models/migrations/core/01-initial/up.sql
@@ -48,14 +48,13 @@ CREATE UNIQUE INDEX path_uidx ON paths(block_group_id, name);
 CREATE TABLE accessions (
   id BLOB PRIMARY KEY NOT NULL,
   name TEXT NOT NULL,
---  path accessions can reference other path accessions
-  path_id BLOB NOT NULL,
+  block_group_id BLOB NOT NULL,
   parent_accession_id BLOB,
-  FOREIGN KEY(path_id) REFERENCES paths(id),
+  FOREIGN KEY(block_group_id) REFERENCES block_groups(id),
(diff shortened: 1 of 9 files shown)


gen-models/migrations/core/01-initial/up.sql
```diff
@@ -48,14 +48,13 @@ CREATE UNIQUE INDEX path_uidx ON paths(block_group_id, name);
 CREATE TABLE accessions (
   id BLOB PRIMARY KEY NOT NULL,
   name TEXT NOT NULL,
---  path accessions can reference other path accessions
-  path_id BLOB NOT NULL,
+  block_group_id BLOB NOT NULL,
```
Chris7, line 51: we are doing this in initial because there isn't a clean migration path that is simple. the hash id relies on path_id, which is now gone, so if we migrated data get_or_create would still fail.

gen-python/src/python_api/jupyter_widget.rs
```diff
@@ -270,8 +271,22 @@ impl PyGraphController {
         conn: &GraphConnection,
         group: &str,
     ) -> Result<AnnotationTrack, AnnotationError> {
+        let block_group_id = self
```
Chris7, line 274: fix bindings from a #125

gen-models/migrations/core/01-initial/up.sql
dkhofer: Yep.  Feel free to squash everything into 01 here if you want

gen-python/src/python_api/jupyter_widget.rs
dkhofer: :+1: Thanks

dkhofer (approved):
LGTM!

PR #139, opened 2026-05-14
Nonblocking opendal calls

dkhofer:


gen-models/src/assets.rs
```diff
@@ -496,7 +411,7 @@ impl LocalAssetUri {
 
     pub fn close_write(&mut self) -> io::Result<()> {
         if let Some(mut file) = self.write_file.take() {
-            file.close()?;
+            file.flush()?;
```
Chris7, line 414: why flush instead of closing?

gen-models/src/assets.rs
```diff
@@ -192,64 +148,21 @@ impl OpenDalLocation {
...
-        let mut writer = asset_location
-            .operator
-            .writer(&asset_location.path)
-            .map_err(io::Error::other)?
-            .into_std_write();
-        let bytes_copied = io::copy(&mut reader, &mut writer)?;
-        writer.close()?;
-        Ok(bytes_copied)
+        Ok(ChecksummedReader::new(Box::new(Cursor::new(
+            reader.to_vec(),
```
Chris7, line 164: this is going to consume the entire reader, correct? We need to stream files otherwise they'll be read into memory.

gen-models/src/assets.rs
```diff
@@ -117,44 +113,6 @@ impl Drop for ChecksummedReader {
 }
 
 impl OpenDalLocation {
-    fn new_fs(root: &Path, path: &Path) -> Result<Self, opendal::Error> {
```
Chris7, line 120: i'm confused why all these fuctions were removed. but tbh i also forgot what they were solving

gen-models/src/assets.rs
```diff
@@ -677,10 +592,7 @@ impl LocalAssetUri {
             return Ok(());
         }
 
-        OpenDalLocation::from_workspace_path(workspace, source_path)
-            .map_err(opendal_file_addition_error)?
-            .copy_to_local_path(workspace, &asset_path)
-            .map_err(FileAdditionError::FileReadError)?;
+        fs::copy(source_path, &asset_path).map_err(FileAdditionError::FileReadError)?;
```
Chris7, line 595: is the opendal fs service a problem to use?

Chris7 (commented):
some questions to get my bearings

gen-models/src/assets.rs
dkhofer: Basically this is getting rid of code that wraps file reads with the opendal runtime.  I'm doing this because on the genhub side, the async code that handles a push is executed within the tokio runtime.  Tokio blows up when it tries to create new file additions and hits the point where the opendal runtime is used.  I'm not sure exactly why they can't run together, but it seems they can't

gen-models/src/assets.rs
dkhofer: Yes.  When code inside a tokio async handler calls into fileaddition code, tokio crashes, I think because it doesn't like the opendal runtime

gen-models/src/assets.rs
dkhofer: This is the end of a new reader_async method, so I think it's doing the right thing.  The other methods (checksum, copy to local path) are getting removed because they're not needed if we just use straight file operations.  

That said, I'm not holding up this PR as an end solution for the problem, I just wanted to remove the opendal runtime stuff to confirm it fixes the problem on the genhub side

gen-models/src/assets.rs
dkhofer: The write_file field is changing from an opendal::blocking::StdWriter to a standard rust File, and the documentation says files are closed automatically when they go out of scope: https://doc.rust-lang.org/std/fs/struct.File.html  I assume StdWriter had a close() method

dkhofer: (There is no close() method for File objects.)

PR #140, opened 2026-05-14
feat(gen-python): make collection optional and last in all PyO3 bindings

bobvh:
## Summary

- Renames `name` → `collection` across all PyO3 repository methods (imports, exports, updates, graph_ops) and moves it to the last parameter position so callers can ignore it in the common case
- Adds `repository_api.ipynb` — a fully-executed example notebook covering imports, exports, updates, and graph operations, with a multi-omics section demonstrating collection-based organisation (genomics / transcriptomics / proteomics in one repo)
- Notebook uses `show_path()` on the edited sample after each update operation to highlight the edited path

### DIFF
--- a/gen-python/examples/repository_api.ipynb
+++ b/gen-python/examples/repository_api.ipynb
@@ -1,0 +1,4740 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "01-header",
+   "metadata": {},
+   "source": [
+    "# Repository API\n\nDemonstrates imports, exports, updates, and graph operations.\nAll functions accept an optional `collection` parameter (see Section 5)."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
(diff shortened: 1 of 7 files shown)


bobvh:
Two issues uncovered while building the example notebook:

**`show_path()` should filter by chromosome index** rather than using the concept of "current path", which doesn't reliably point to the edited path after an update.

**`parent_samples` should be renamed `reference_graphs`** - the argument doesn't require an actual sample, just a reference graph, so the name is misleading.

bobvh:
> Two issues uncovered while building the example notebook:
> 
> **`show_path()` should filter by chromosome index** rather than using the concept of "current path", which doesn't reliably point to the edited path after an update.
> 
> **`parent_samples` should be renamed `reference_graphs`** - the argument doesn't require an actual sample, just a reference graph, so the name is misleading.

Path issue is resolved by PR143, parent samples is fixed in main

PR #143, opened 2026-05-15
(viewers) indicate which edges are real vs history

bobvh:
Uses chromosome index and the pruning logic to make it clear which edges represent the wildtype vs edit. Without this, you can't tell apart a heterozygous from homozygous mutant.


gen-tui/src/graph_controller.rs
```diff
@@ -69,6 +69,12 @@ where
     /// Persistent requested highlights in domain terms
     pub highlights: Vec<(HighlightKind<G::NodeId>, PathStyle)>,
 
+    /// Persistent dimmed edges in domain terms: (src_id, tgt_id)
```
bobvh, line 72: What's different about "lowlights" and why they're not just a different style highlight is that highlights dominate any other edge that uses the same routing segment, while for lowlights it's the opposite: you only see them marked if it's only those edges that make use of the segment.

This extends to how we handle the junctions: with highlights the bold corners could form a neat overlay, whereas for lowlights we want them to fade into the background.

src/views/snapshots/r#gen__views__gen_graph_widget__tests__zygosity_pruned_edges.snap
```diff
@@ -0,0 +1,34 @@
...
+"                                                                                                                        "
+"                                                                                                                        "
+"                                                                                                                        "
+"                                                                                                                        "
+"                                                                                                                        "
+"                                                                                                                        "
+"          ╭─T─╮     ╭─A─╮      ╭╌G╌╮      ╭─A─╮       ╭─TTT─╮       ╭─A─╮                                               "
+"     ╟─AT─╯   ╰─GAT─┤   ├─GATC─╮   ╭─ATCG─┤   ├─TCGGG─╯     ╰─AACAC─┤   ├─CAGAGA─╢                                      "
+"          ╰╌C╌╯     ╰─C─╯      ╰───╯      ╰───╯       ╰╌╌╌╌╌╯       ╰─G─╯                                               "
```
bobvh, line 20: This looks a lot more clear in my terminal than on github

src/views/gen_graph_widget.rs
```diff
@@ -188,6 +191,83 @@ pub fn create_gen_graph_widget(
     GraphWidget::with_renderer(renderer)
 }
 
+/// Compute which edges would be removed by `BlockGroup::prune_graph`.
```
bobvh, line 194: @Chris7 this is where review would be most useful. Does this capture all the cases of edges we want to fade/hide?

src/views/snapshots/r#gen__views__gen_graph_widget__tests__zygosity_pruned_edges.snap
bobvh: Rendered in the actual widget (color + dashed lines): 
image

bobvh: This is the VCF corresponding to that:
```
##fileformat=VCFv4.2
##reference=simple.fa
##contig=<ID=m123,length=34>
##FORMAT=<ID=GT,Number=1,Type=String,Description="Genotype">
#CHROM	POS	ID	REF	ALT	QUAL	FILTER	INFO	FORMAT	SAMPLE1
m123	3	.	C	T	60	PASS	.	GT	1|1
m123	7	.	C	A	60	PASS	.	GT	0|1
m123	11	.	CG	C	60	PASS	.	GT	1|1
m123	16	.	GA	G	60	PASS	.	GT	0|1
m123	22	.	A	ATTT	60	PASS	.	GT	1|1
m123	28	.	C	G	60	PASS	.	GT	0|1
```

Looks correct: homozygous alt, heterozygous, homozygous alt, heterozygous, homozygous alt, heterozygous

gen-tui/src/plotter.rs
```diff
@@ -301,12 +305,24 @@ pub fn plot_viewport_graph_with_highlights<R, G>(
             .map(|(_, style)| *style)
             .next_back();
 
+        let is_lowlight = lowlights
```
Chris7, line 308: Can lowlights be a hashmap so these can be lookups. Is there a reason it's an array

gen-tui/src/plotter.rs
```diff
@@ -464,12 +509,51 @@ pub fn plot_viewport_graph_with_highlights<R, G>(
     }
 }
 
-/// Compute the junction glyph for a routing node based on its connections
-fn compute_junction_glyph(viewport_graph: &CroppedGraph, pos: WorldPos) -> JunctionSymbol {
-    let mut connections = 0u8;
+/// Returns true if both `point_a` and `point_b` lie on the axis-aligned segment
```
Chris7, line 512: It's more useful to say why we do something and it's goal vs statements we can read from the code

gen-tui/src/plotter.rs
```diff
@@ -499,7 +582,7 @@ fn draw_edge_with_style(
     let (v_ch, h_ch) = match line_style {
         LineStyle::Normal => ('│', '─'),
         LineStyle::Bold => ('┃', '━'),
-        LineStyle::Dashed => ('┆', '┄'),
+        LineStyle::Dashed => ('┊', '╌'),
```
Chris7, line 585: You should make these static consts at the global level and define them

src/views/gen_graph_widget.rs
Chris7: If we're reusing the prune logic, it might be better to make that a method to find which ones to prune and use it in both places than reimplement it here

gen-tui/src/graph_controller.rs
```diff
@@ -149,6 +155,8 @@ where
             rebuild_needed: true,
             layout_changed: true, // Treat initial build as a layout change to place the camera
             highlights: Vec::new(),
+            edge_lowlights: Vec::new(),
```
Chris7, line 158: seems like a hashmap here would make it easier to know when to lowlight/hilight

gen-tui/src/graph_controller.rs
```diff
@@ -487,6 +495,85 @@ where
...
+        let lowlight_pairs: Vec<(NodeIndex, NodeIndex)> = domain_lowlights
+            .iter()
+            .map(|(src, tgt)| {
+                let u = NodeIndex::new(<G as NodeIndexable>::to_index(graph, *src));
+                let v = NodeIndex::new(<G as NodeIndexable>::to_index(graph, *tgt));
+                (u, v)
+            })
+            .collect();
+
+        let edges: Vec<(WorldPos, WorldPos)> = viewport_graph
```
Chris7, line 529: use a hashmap, we have nested for loops

src/views/gen_graph_widget.rs
```diff
@@ -188,6 +191,83 @@ pub fn create_gen_graph_widget(
...
+    }
+
+    pruned
+}
+
+/// Find nodes that become inaccessible when all pruned edges are removed.
+///
+/// BFS from all start nodes following only non-pruned edges. Any node not reached
+/// is only reachable through pruned (lowlighted) edges and should be dimmed.
+fn compute_inaccessible_nodes(
```
Chris7, line 249: i think this exists in gen-graph already as orphan nodes or something

Chris7 (approved):
The prune looks correct. If it's a copy/paste from the other code, it should be refactored into a method both can use. Some other places have places that seem inefficient and don't have to be.

PR #144, opened 2026-05-18
Extract GraphLocus and BlockSlice into gen-models::locus

bobvh:
Extracts `GraphLocus` and `BlockSlice` from `graph_search` into a shared
`gen-models::locus` module. GraphLocus was also rewritten to use
a list of slices from GraphNodes (Blocks) instead of a list of Blocks
and two offsets.

To convert between GraphLocus (Block graph) and AnnotationSpan (Node graph)
there's annotation_span_from_graph_locus and graph_locus_from_annotation_span
in src/views/annotation_track.rs.


.gitignore
```diff
@@ -34,6 +34,9 @@ rust-analyzer.json
 # Terminal recording via VHS
 *.tape
 
+# R bindings (library crate — lock not committed)
```
bobvh, line 37: Internet says that cargo.lock should only be committed for applications, not libraries. Do we have a different policy internally?

gen-models/src/locus.rs
```diff
@@ -0,0 +1,78 @@
...
+use gen_graph::GraphNode;
+use serde::{Deserialize, Serialize};
+
+use crate::{db::GraphConnection, node::Node, sequence::reverse_complement};
+
+/// A slice of a single graph block: the block itself plus local start/end byte
+/// offsets within that block's sequence.  Middle blocks in a multi-block locus
+/// span the full block (`start = 0`, `end = block.length()`).
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct BlockSlice {
```
Chris7, line 28: Seems better to call these nodeslices instead and put them in gen-graph

gen-models/src/sequence.rs
```diff
@@ -507,6 +507,31 @@ impl Query for Sequence {
     }
 }
 
+pub fn reverse_complement(seq: &[u8]) -> Vec<u8> {
```
Chris7, line 510: I could swear this exists already but I'm on mobile and can't check

Chris7 (approved):
If it doesn't require new imports, I think your graphslice should be nodeslices and in gen-graph. The part using SQL should stay in gen-models

gen-models/src/locus.rs
```diff
@@ -0,0 +1,78 @@
...
+            end: block.length() as usize,
+        }
+    }
+}
+
+/// A region in graph space expressed as an ordered list of block slices.
+/// The strand field specifies the 5'-3' orientation for double stranded DNA.
+///
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct GraphLocus {
```
Chris7, line 50: this wouldn't support trans splicing by having all features share a strand

gen-models/src/locus.rs
bobvh: technically they're graphnodeslice then if we follow gengraph standard. The coordinates that define them are relative to the start of the graphnode, so I didn't want to use node. I thought graphnodes are also called blocks?

I'll move it to gen-graph/src/lib.rs , but we should pick up that old renaming issue too.

gen-models/src/sequence.rs
bobvh: I moved it from graph_search.rs into sequence.rs, that must be the one you remember (introduced in 893dbfb1)

gen-models/src/locus.rs
bobvh: Would having strand on each individual slice cover it then? Or is this something where the edges come in?

PR #145, opened 2026-05-18
Use AnnotationSpan for inline annotations

bobvh:
Make inline annotations (not carried on tracks) use annotationspan instead of GraphLocus so that the inline vs tracked rendering more closely resembles each other and can use common infrastructure.


bobvh:
Failed tests were the result of a last minute rebase to move a commit to its logical spot. Test run fine locally now and i've updated my hooks to make this not happen in the future.

PR #146, opened 2026-05-19
Add method to merge operation file info with file addition

Chris7:
This provides a query helper for getting the union of information from operation_files + file addition table

### DIFF
--- a/gen-models/src/operations.rs
+++ b/gen-models/src/operations.rs
@@ -371,6 +371,29 @@ pub struct OperationFile {
     pub checksum_override: Option<HashId>,
 }
 
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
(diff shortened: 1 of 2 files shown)


gen-models/src/operations.rs
```diff
@@ -397,6 +420,53 @@ impl OperationFile {
         self.checksum_override = Some(checksum);
         self
     }
+
+    pub fn get_files_for_operation(
+        conn: &OperationsConnection,
+        operation_hash: &HashId,
+    ) -> Vec<OperationFileInfo> {
```
dkhofer: It'd be good to have this return a Result

dkhofer (approved):
One comment, otherwise LGTM

gen-models/src/operations.rs
Chris7: good catch

PR #147, opened 2026-05-19
Add file path to operation files

Chris7:
Stores the filepath in operation files and always makes fileaddition asset_uri go under .gen/assets if it's a file uri. Previously, asset_uri could be like fasta/hg19.fa, which would not be stable across operations.

### DIFF
--- a/gen-models/migrations/operations/05-operation-files-file-path/down.sql
+++ b/gen-models/migrations/operations/05-operation-files-file-path/down.sql
@@ -1,0 +1,16 @@
+UPDATE file_additions
+SET asset_uri = COALESCE(
+  (
+    SELECT 'file://' || of.file_path
+    FROM operation_files of
+    WHERE of.file_addition_id = file_additions.id
+      AND of.file_path NOT LIKE '%://%'
+    ORDER BY of.id
+    LIMIT 1
+  ),
+  asset_uri
+)
+WHERE asset_uri LIKE 'file://.gen/assets/%';
+
+ALTER TABLE operation_files
+DROP COLUMN file_path;
--- a/gen-models/src/assets.rs
+++ b/gen-models/src/assets.rs
@@ -492,6 +492,42 @@ impl LocalAssetUri {
         Self::stored_relative_path(workspace, &source_path, checksum, file_type)
     }
 
+    pub fn is_asset_relative_path(
+        workspace: &Workspace,
+        relative_path: &str,
+    ) -> Result<bool, FileAdditionError> {
+        let repo_root = workspace.repo_root()?;
+        let asset_dir = workspace.asset_dir()?;
+        let asset_relative_dir =
+            asset_dir
+                .strip_prefix(&repo_root)
+                .map_err(|_| FileAdditionError::PathOutsideRepo {
+                    path: asset_dir.clone(),
+                    repo_root: repo_root.clone(),
+                })?;
+
+        let candidate = if relative_path.is_empty() {
+            PathBuf::new()
+        } else {
+            Self::sanitize_relative_path(Path::new(relative_path))?
+        };
+
+        Ok(candidate.starts_with(asset_relative_dir))
+    }
+
+    pub fn repo_relative_destination_path(
+        workspace: &Workspace,
+        relative_path: &str,
+    ) -> Result<PathBuf, FileAdditionError> {
+        let sanitized = if relative_path.is_empty() {
+            PathBuf::new()
+        } else {
+            Self::sanitize_relative_path(Path::new(relative_path))?
+        };
+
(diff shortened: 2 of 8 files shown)


gen-models/migrations/operations/05-operation-files-file-path/down.sql
```diff
@@ -0,0 +1,16 @@
+UPDATE file_additions
```
Chris7, line 1: going to squash soon

gen-models/src/assets.rs
```diff
@@ -475,6 +478,56 @@ impl LocalAssetUri {
...
+    ) -> Result<String, FileAdditionError> {
+        let source_path = if Self::is_file_uri(path_or_uri) {
+            Self::resolve_source_path(workspace, path_or_uri)?
+        } else {
+            Self::resolve_input_source_path(workspace, path_or_uri)?
+        };
+        Self::stored_relative_path(workspace, &source_path, checksum, file_type)
+    }
+
+    pub fn is_asset_relative_path(
```
Chris7, line 496: this checks whether a relative path is in the .gen/assets directory of a workspace

gen-models/src/assets.rs
```diff
@@ -475,6 +478,56 @@ impl LocalAssetUri {
...
+        let candidate = if relative_path.is_empty() {
+            PathBuf::new()
+        } else {
+            Self::sanitize_relative_path(Path::new(relative_path))?
+        };
+
+        Ok(candidate.starts_with(asset_relative_dir))
+    }
+
+    pub fn repo_relative_destination_path(
```
Chris7, line 521: given a workspace + relative_path, try to safely give a desination path

gen-models/src/assets.rs
```diff
@@ -515,6 +568,13 @@ impl LocalAssetUri {
         workspace: &Workspace,
         path_or_uri: &str,
     ) -> Result<PathBuf, FileAdditionError> {
+        if !Self::is_local_path_or_file_uri(path_or_uri) {
```
Chris7: this is a guard from calling this method without a localfileuri

src/operation_management.rs
```diff
@@ -567,6 +568,40 @@ fn apply_operations_to_remote(
...
+                // This is materializing the file in the directory. If file_addition file_path
+                // will always be the asset_dir path, while file_addition.file_path can be like
+                // fastas/hg19.fa. So if we are not in the asset_dir, we copy it out
+                if file_addition.file_path != file_addition.file_addition.file_path()
+                    && LocalAssetUri::is_asset_relative_path(
+                        remote_workspace,
+                        file_addition.file_addition.file_path(),
+                    )?
+                {
+                    let user_dst_path = LocalAssetUri::repo_relative_destination_path(
```
Chris7, line 581: this is the reason for those added functions around checking local assets/paths. It's to try and prevent someone from making a ManifestOperation with a filepath of like /etc/passwd

src/operation_management.rs
```diff
@@ -567,6 +568,40 @@ fn apply_operations_to_remote(
...
+                        remote_workspace,
+                        &file_addition.file_path,
+                    )
+                    .map_err(|err| RemoteOperationError::IOError(std::io::Error::other(err)))?;
+                    if user_dst_path.exists() {
+                        info!(
```
Chris7, line 587: at some point i guess we want to decide if we should overwrite.

gen-models/migrations/operations/05-operation-files-file-path/down.sql
dkhofer: Good

gen-models/src/assets.rs
dkhofer: :+1:

dkhofer: That could be a documentation comment for this function

gen-models/src/assets.rs
dkhofer: This could also be a comment for the function

gen-models/src/assets.rs
dkhofer: It'd be good to name the variable `file_path_or_uri` to make it clear that's what's expected

src/operation_management.rs
dkhofer: :+1:

src/operation_management.rs
dkhofer: Yep

dkhofer (approved):
LGTM, just a few minor comments

gen-models/src/assets.rs
Chris7: good idea

PR #148, opened 2026-05-20
ci: add notebook test workflow using pytest --nbmake

bobvh:
Runs `pytest --nbmake` with a 120s timeout per cell on every ipynb file in the repo. 
`#noqa` skips a single cell, if we want to ignore entire files a conftest.py at project root with collect_ignore should do the trick. Currently some of the legacy notebooks are extremely slow. I'm leaning more towards fixing those and excluding cell by cell if needed.


Chris7:
Your new workflow is failing

bobvh:
> Your new workflow is failing

Yes, it's old notebooks that either take too long & download a bunch of data, or have deprecated function calls. I'll fix the latter, and will make an ignore file for the the former.

Chris7:
no ,it's failing to install the requirements. It never gets to testing

bobvh:
The conftest.py currently excludes all notebooks under docs/ and examples/, which is broader than ideal. As more tested notebooks are added to gen-python/examples/, we should revisit and narrow these exclusions.

.claude/scheduled_tasks.lock
```diff
@@ -0,0 +1 @@
+{"sessionId":"3f097e44-1727-46e6-872e-12a12d270f9a","pid":8222,"acquiredAt":1779645045694}
```
Chris7: What is this

.github/workflows/test-notebooks.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Build and install gen package
+        run: |
+          python -m venv .venv
+          source .venv/bin/activate
+          pip install maturin
+          maturin develop
+      - name: Install notebook test dependencies
+        run: source .venv/bin/activate && pip install pytest nbmake
+      - name: Test notebooks
+        run: source .venv/bin/activate && pytest --nbmake --nbmake-timeout=120 gen-python/examples/
```
Chris7: Can the notebooks be fixed so this merges in as passing

bobvh:
Not supposed to end up in git, I was using the remote interface on my phone
so I didn’t see the commit (I wasn’t going to merge like that to be clear).
The scheduled task is claude looking at the results of the CI job after
giving it 2 minutes to run or fail.

.github/workflows/test-notebooks.yml
bobvh: Done

.claude/scheduled_tasks.lock
bobvh: See reply in OP (timer so that it doesn't wait for CI to finish but instead checks on it after 2.5 min)

bobvh:
@Chris7 CI is all green now, including `test-notebooks`, without ignoring the legacy notebooks.

PR #151, opened 2026-05-21
feat(jupyter-widget): per-cell cloning, freeze, and snapshots

bobvh:
## Summary
- Add per-cell widget cloning so each Jupyter cell gets its own
  independent interactive graph instance
- Add freeze button to lock a widget as a static image
- Auto-snapshot with debounce sends PNG to Python on each frame change;
  freeze and snapshot use separate message types
- Fix window listener memory leaks on widget destroy
- Inset canvas border via `box-shadow` so toggling editable state
  causes no layout shift; swap canvas→img inside wrapper on freeze
  to reduce reflow jitter in content below

### DIFF
--- a/gen-python/js/grid.ts
+++ b/gen-python/js/grid.ts
@@ -30,6 +30,8 @@ export interface GridMetrics {
   boxDrawY: number;
   textX: number;
   textBaseline: number;
+  textSize: number;
+  cellSize: number;
 }
 
 export function cellFont(size: number, cell: Pick<Cell, "bold" | "italic">): string {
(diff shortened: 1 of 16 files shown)


gen-python/js/grid.ts
```diff
@@ -54,19 +56,23 @@ export function isBoxLike(ch: string): boolean {
   );
 }
 
-export function makeGridMetrics(ctx: CanvasRenderingContext2D): GridMetrics {
-  const block = measure(ctx, BOX_FONT, "\u2588");
+export function makeGridMetrics(ctx: CanvasRenderingContext2D, scale = 1): GridMetrics {
+  const textSize = TEXT_SIZE * scale;
```
bobvh, line 60: I'm doing this because the PNG snapshot of the canvas otherwise is at half the actual resolution (sub-pixels).

gen-python/js/index.ts
```diff
@@ -44,20 +44,42 @@ function render({ model, el }: RenderContext): void {
 
   const btnContainer = document.createElement("div");
   btnContainer.style.cssText =
-    "position: absolute; bottom: 8px; right: 8px; display: flex; flex-direction: column; gap: 4px; z-index: 1;";
+    "position: absolute; top: 8px; right: 8px; display: flex; flex-direction: row; gap: 4px; z-index: 1;";
 
   const zoomInBtn = document.createElement("button");
-  zoomInBtn.textContent = "+";
+  zoomInBtn.innerHTML = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="18" height="18" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round">
```
bobvh: The plus symbol alone could be misinterpreted.

gen-python/js/index.ts
```diff
@@ -66,35 +88,67 @@ function render({ model, el }: RenderContext): void {
   let nodeCells = new Set<string>();
   let frozen = false;
 
+  let snapshotTimer: ReturnType<typeof setTimeout> | null = null;
```
bobvh: Wait 1 second after the last update before taking the snapshot to store in the notebook.

PR #153, opened 2026-05-22
Update block group methods to return results as needed

dkhofer:
I noticed with the recent changes involving intervaltree, some BlockGroup methods should be returning Results but aren't.  This addresses that

### DIFF
--- a/gen-models/src/block_group.rs
+++ b/gen-models/src/block_group.rs
@@ -336,7 +336,12 @@ impl BlockGroup {
         }
     }
 
-    pub fn delete(conn: &GraphConnection, collection_name: &str, sample_name: &str, name: &str) {
+    pub fn delete(
+        conn: &GraphConnection,
+        collection_name: &str,
+        sample_name: &str,
+        name: &str,
+    ) -> Result<(), BlockGroupError> {
         let query = "delete from block_groups where collection_name = ?1 and sample_name = ?2 and name = ?3;";
         conn.execute(
             query,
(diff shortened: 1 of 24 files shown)


gen-models/src/block_group.rs
```diff
@@ -559,11 +562,14 @@ impl BlockGroup {
...
-    pub fn get_graph(conn: &GraphConnection, block_group_id: &HashId) -> GenGraph {
+    pub fn get_graph(
+        conn: &GraphConnection,
+        block_group_id: &HashId,
+    ) -> Result<GenGraph, BlockGroupError> {
         let edges = BlockGroupEdge::edges_for_block_group(conn, block_group_id);
         let blocks = Edge::blocks_from_edges(conn, &edges);
         let (graph, _) = Edge::build_graph(&edges, &blocks);
-        graph
+        Ok(graph)
```
dkhofer, line 572: This is a little dumb, but it does make it easier to change the query methods in this method to return results later

dkhofer: I'm not doing that here because this PR is big enough already

gen-models/src/traits.rs
```diff
@@ -33,7 +35,21 @@ pub trait Query {
         objs
     }
 
-    fn get(conn: &Connection, query: &str, params: impl Params) -> Result<Self::Model> {
+    fn try_query(
```
dkhofer, line 38: This is just the same as query() but returns a Result.  If that seems reasonable we can work on changing other things over to try_query (or change query to return a Result)

Chris7: i say make query a result

Chris7 (approved):
lgtm. I think we should make query a result

gen-models/src/traits.rs
dkhofer: I'm going to do that in a followup PR

PR #154, opened 2026-05-24
Enforce relative paths for end_operation

Chris7:
end_operation wasn't normalizing file paths. I updated add_file to enforce relative paths there so there's fewer places where paths we don't want sneak in.

### DIFF
--- a/gen-models/src/block_group.rs
+++ b/gen-models/src/block_group.rs
@@ -3237,7 +3237,7 @@ mod tests {
     #[test]
     fn test_blockgroup_interval_tree() {
         let conn = &get_connection(None).unwrap();
-        let (block_group_id, path) = setup_block_group(conn);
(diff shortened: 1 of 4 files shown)


gen-models/src/block_group.rs
```diff
@@ -3237,7 +3237,7 @@ mod tests {
     #[test]
     fn test_blockgroup_interval_tree() {
         let conn = &get_connection(None).unwrap();
-        let (block_group_id, path) = setup_block_group(conn);
+        let (block_group_id, _path) = setup_block_group(conn);
```
Chris7, line 3240: clippy

dkhofer: :+1:

dkhofer (approved):
LGTM

PR #155, opened 2026-05-26
Address PR144 remarks

bobvh:
Jumped the gun on merging PR144: https://github.com/genhub-bio/gen/pull/144

### DIFF
--- a/gen-graph/src/lib.rs
+++ b/gen-graph/src/lib.rs
@@ -59,14 +59,17 @@ pub struct BlockSlice {
     pub start: usize,
     /// Local end offset, exclusive (`start..=block.length()`).
     pub end: usize,
+    /// 5'→3' orientation of this slice.
+    pub strand: Strand,
 }
 
 impl BlockSlice {
-    pub fn full(block: GraphNode) -> Self {
+    pub fn full(block: GraphNode, strand: Strand) -> Self {
         Self {
             block,
             start: 0,
             end: block.length() as usize,
+            strand,
         }
     }
 }
--- a/gen-graph/src/lib.rs
+++ b/gen-graph/src/lib.rs
@@ -49,6 +49,28 @@ impl GraphNode {
     }
 }
 
+/// A slice of a single graph block: the block itself plus local start/end byte
+/// offsets within that block's sequence.  Middle blocks in a multi-block locus
+/// span the full block (`start = 0`, `end = block.length()`).
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct BlockSlice {
+    pub block: GraphNode,
+    /// Local start offset within the block's sequence slice (`0..block.length()`).
+    pub start: usize,
+    /// Local end offset, exclusive (`start..=block.length()`).
+    pub end: usize,
+}
+
+impl BlockSlice {
+    pub fn full(block: GraphNode) -> Self {
+        Self {
+            block,
+            start: 0,
+            end: block.length() as usize,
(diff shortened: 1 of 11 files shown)


gen-graph/src/lib.rs
```diff
@@ -49,6 +49,31 @@ impl GraphNode {
     }
 }
 
+/// A slice of a single graph block: the block itself plus local start/end byte
+/// offsets within that block's sequence.  Middle blocks in a multi-block locus
+/// span the full block (`start = 0`, `end = block.length()`).
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+pub struct BlockSlice {
```
bobvh: This addresses https://github.com/genhub-bio/gen/pull/144#pullrequestreview-4351289780 : definition should live in gen-graph instead of gen-models. I didn't change the name yet because I think that NodeSlice would muddy the water between Graphnode and Node more. I was using Block in relation to "blockgroup", but can see a few scenarios in which this does become NodeSlice.

(For example if we do:
- Node becomes Supernode as a stable, never-split sequence unit; "super" encodes that nodes live inside it
- GraphNode becomes Node as the thing edges really connect to, following conventional pangenome terms and hopefully confusing LLMs less
  - BlockSlice becomes NodeSlice as a coordinate-bounded slice of a Node
  - BlockGroup becomes SequenceGraph as discussed earlier.

bobvh:
I confirmed that the reverse_complement method doesn't occur anywhere else in the repo, PR144 moved it from graph_search to sequence.rs.

gen-python/src/python_api/graph_search.rs
```diff
@@ -92,49 +92,26 @@ impl PyGraphLocus {
         PyGraphPos::new(s.block, s.end)
     }
 
-    /// Ordered sequence of blocks that span this locus.
+    /// Ordered sequence of block slices that span this locus.
```
bobvh, line 95: Python was still on the old representation (blocks and offsets rather than the slices that are more akin to anonymous annotationspans)

gen-r/src/rust/src/lib.rs
```diff
@@ -179,6 +179,27 @@ fn block_record(node_id: HashId, sequence_start: i64, sequence_end: i64) -> Robj
     obj
 }
 
+fn block_slice_record(
```
bobvh, line 182: Like Python, R was on the old GraphLocus representation still. This fixes that too

gen-models/src/locus.rs
```diff
@@ -16,40 +16,16 @@
...
-
 /// A region in graph space expressed as an ordered list of block slices.
-/// The strand field specifies the 5'-3' orientation for double stranded DNA.
-///
+/// Each slice carries its own strand, allowing trans-spliced loci where
+/// individual exons may come from opposite strands.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct GraphLocus {
     pub slices: Vec<BlockSlice>,
-    pub strand: Strand,
```
bobvh, line 52: This addresses annotations made from loci not being able to support [certain forms of splicing](https://github.com/genhub-bio/gen/pull/144#discussion_r3293439014).

gen-graph/src/lib.rs
Chris7: Keep it GraphNodeSlice for now. It's slicing a GraphNode. If we rename stuff later it'll make it simpler

Chris7 (approved):
Only comment is on the name. My preference is being consistent with what exists now.

PR #158, opened 2026-05-28
fix(search): seed-index search includes reverse-complement hits

bobvh:
- `find_all_with_seed_index` was only searching the forward orientation; DNA-mode searches now also search the reverse complement, matching the behavior of `find_all`
- RC hits are tagged `Strand::Reverse`; forward hits are tagged `Strand::Forward`
- Short queries (< k) fall back to a full scan and also include RC hits

New tests: `seed_index_finds_reverse_complement_via_index`, `seed_index_short_query_finds_reverse_complement`

### DIFF
--- a/src/graphs/graph_search.rs
+++ b/src/graphs/graph_search.rs
@@ -274,12 +274,12 @@ impl GenGraphMatcher {
         out
     }
 
-    /// Returns every forward-only match using a seed index to prune start positions.
+    /// Returns every match of `query` using a seed index to prune start positions.
     ///
-    /// Does not perform reverse-complement search. Does not expand IUPAC-degenerate
-    /// bases. Returns an error if the query contains degenerate bases, because an
-    /// exact seed lookup would silently miss valid matches.
-    ///
+    /// For `SequenceKind::Dna`, also searches the reverse complement and tags
+    /// results with `Strand::Forward` or `Strand::Reverse`. Does not expand
+    /// IUPAC-degenerate bases. Returns an error if the query contains degenerate
+    /// bases, because an exact seed lookup would silently miss valid matches.
     pub fn find_all_with_seed_index(
         &self,
         seed_index: &SeedIndex,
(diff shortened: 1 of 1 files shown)


src/graphs/graph_search.rs
```diff
@@ -765,6 +795,15 @@ mod tests {
         GenGraphMatcher::new(conn, graph)
     }
 
+    fn test_ssdna_matcher() -> GenGraphMatcher {
+        let ctx = setup_gen();
+        let conn = ctx.graph().conn();
+        let _ = Collection::create(conn, "test");
+        let (block_group_id, _path) = setup_block_group(conn);
+        let graph = BlockGroup::get_graph(conn, &block_group_id);
+        GenGraphMatcher::new_ssdna(conn, graph)
```
dkhofer, line 797: I'm confused, nothing is being asserted here.  Is the test just confirming that an ssDNA matcher can be constructed?

dkhofer: Oh, never mind, this isn't an actual test.  It would be good to rename this function to build_ssdna_matcher

src/graphs/graph_search.rs
```diff
@@ -1019,6 +1058,54 @@ mod tests {
         assert_eq!(err, SeedIndexSearchError::UnsupportedQuery);
     }
 
+    #[test]
+    fn seed_index_finds_reverse_complement_via_index() {
+        let matcher = test_matcher();
+        let index = SeedIndex::build(&matcher, 4, true);
+
+        // The graph contains AAAAAAAAAA (A node) and TTTTTTTTTT (T node).
```
dkhofer, line 1059: Thanks for the comments like this

src/graphs/graph_search.rs
```diff
@@ -299,8 +299,35 @@ impl GenGraphMatcher {
             |q: u8, g: u8| q == g
         };
 
+        let mut out = self.index_search_one_orientation(seed_index, query, matcher);
+
+        if self.sequence_kind == SequenceKind::Dna {
+            for m in out.iter_mut() {
```
dkhofer: FYI, modifying things in a list in place is usually an antipattern.  The main reason to do it is to avoid allocating memory (to make things faster), which often isn't important these days.

Since the strand is known when searching (or you could at least default to unknown), why not pass that in as a parameter to the search, so that it can create GraphNodeSlices with that strand value?

dkhofer (commented):
Mostly looks good, but I think handling the strand could be improved

src/graphs/graph_search.rs
bobvh: Fixed, thanks!

PR #165, opened 2026-05-29
Graph Traversal v2

Chris7:


Makefile
```diff
@@ -53,7 +53,7 @@ clean:
 build:
 	cargo build --all-features
 clippy-fix:
-	cargo clippy --all-targets --all-features --allow-dirty --allow-staged --fix -- -D clippy::all
+	cargo clippy --all-targets --all-features --allow-dirty --allow-staged --fix -- -D clippy::all || cargo +nightly fmt --all
```
bobvh, line 56: do we have to get on nightly builds from cargo for this?

gen-graph/src/lib.rs
```diff
@@ -792,6 +880,362 @@ mod tests {
...
+
+        let d_in: Vec<_> = graph
+            .neighbors_directed(node_d, petgraph::Direction::Incoming)
+            .collect();
+        assert_eq!(d_in, vec![node_c]);
+    }
+
+    #[test]
+    fn test_graph_from_interval_tree_with_fragments() {
+        // Nodes can be fragments: different (seq_start, seq_end) for the same node_id
```
bobvh, line 1080: I finally get it: NodeIntervalBlocks are GraphNodes that hold two extra coordinates, referring to a single linear reference...

bobvh:
Looks good. I get how the lazy loading works, I like the multiple position concept (and will be implementing  __add__ and __sub__ on PyAnnotation too). I still don't quite get why you need to carry around the coordinates in relation to an original linear reference, but I'll assume it's a consequence of the original path-based infrastructure and a nice shortcut if you're working with GFFs on that same linear reference.

bobvh (approved):
LGTM and helpful to see how NodeIntervalBlocks work. Do you think we'll have something similar on GenGraph too, as the non-path-linked analog?

gen-graph/src/lib.rs
```diff
@@ -87,6 +88,20 @@ pub struct GraphEdge {
     pub created_on: i64,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Ord, PartialOrd, Deserialize, Serialize)]
+pub struct GraphNodePosition {
```
Chris7, line 93: should we call this GraphNodeCursor?

gen-models/src/region.rs
```diff
@@ -249,7 +253,7 @@ fn resolve_target(
         RegionTargetKind::Path | RegionTargetKind::BlockGroup => {
             start < 0 || end > target.feature_length
         }
-        RegionTargetKind::Annotation | RegionTargetKind::Accession => end > target.feature_length,
+        RegionTargetKind::Annotation | RegionTargetKind::Accession => false,
```
Chris7, line 256: no out of bounds for accessions

gen-graph/src/lib.rs
bobvh: In my mental model I had Position meaning "a location that is exactly one nucleotide wide (i.e. defined by one coordinate)" like here, and Cursor as name for the Vec<Pos>: you pass it around as one object that's often just one position, but sometimes a superposition of multiple possibilities. (I'll update the struct in the viewer to handle it and/or avoid collision)

gen-graph/src/lib.rs
```diff
@@ -395,22 +412,95 @@ pub fn flatten_to_interval_tree(
         }
     }
 
+    let total_length = spans
```
dkhofer: I think `max_endpoint` or `max_block_end` would be better names than `total_length`.  This isn't a length of anything

gen-graph/src/lib.rs
```diff
@@ -395,22 +412,95 @@ pub fn flatten_to_interval_tree(
...
+            let block = normalize_sentinel_block(block, total_length);
+            (block.start..block.end, block)
         })
         .collect();
     tree
 }
 
+pub fn graph_from_interval_tree(tree: &IntervalTree<i64, NodeIntervalBlock>) -> GenGraph {
+    let mut graph = GenGraph::new();
+    let total_length = tree
```
dkhofer: Same here

gen-graph/src/lib.rs
dkhofer: I think GraphNodePosition is more accurate

dkhofer: Based on the actual usage of this struct

gen-graph/src/lib.rs
```diff
@@ -792,6 +882,362 @@ mod tests {
...
+            sequence_end: 5,
+        };
+
+        assert_eq!(node_b.length(), 2);
+        assert!(graph.contains_edge(node_a, node_b));
+        assert!(graph.contains_edge(node_b, node_c));
+    }
+
+    #[test]
+    fn test_graph_from_interval_tree_keeps_sentinels() {
```
dkhofer: Maybe sentinels should be called terminals?  How much do we care about distinguishing start/end nodes at the start/end of a graph from start/end nodes that connect to internal parts of a graph?

gen-models/src/graph.rs
```diff
@@ -0,0 +1,661 @@
...
+
+    let mut neighbor_ids: Vec<HashId> = edges_1hop
+        .iter()
+        .flat_map(|ae| [ae.edge.source_node_id, ae.edge.target_node_id])
+        .filter(|id| *id != node_id)
+        .collect();
+    neighbor_ids.sort();
+    neighbor_ids.dedup();
+
+    let mut all_edges = edges_1hop;
```
dkhofer: It'd be slightly better to make this a hashmap keyed by edge id, then use entry...or_insert to add to it

gen-models/src/graph.rs
```diff
@@ -0,0 +1,661 @@
...
+    }
+}
+
+fn find_offset_with_optional_expansion(
+    graph: &mut GenGraph,
+    anchor: &GraphNodePosition,
+    distance: i64,
+    expand_through_existing_paths: bool,
+    expand: &mut impl FnMut(&mut GenGraph, HashId) -> bool,
+) -> Result<Vec<GraphNodePosition>, GraphError> {
```
dkhofer, line 104: Please add a comment saying what this method does

dkhofer: In particular, what does distance represent

gen-models/src/graph.rs
```diff
@@ -0,0 +1,661 @@
...
+            if remaining <= in_node {
+                results.push(GraphNodePosition {
+                    graph_node: node,
+                    offset: pos.offset + remaining,
+                });
+                continue;
+            }
+
+            let mut neighbors: Vec<GraphNode> = graph
+                .neighbors_directed(node, Direction::Outgoing)
```
dkhofer, line 140: The other thing I was having trouble understanding was outgoing vs. incoming.  It'd be nice to add a comment saying incoming == edges from upstream and outgoing == edges going downstream

gen-models/src/graph.rs
```diff
@@ -0,0 +1,661 @@
...
+
+    Ok(results)
+}
+
+impl ResolvedGraph {
+    pub fn resolve_anchor(
+        &self,
+        coord: i64,
+        conn: &GraphConnection,
+    ) -> Result<GraphNodePosition, GraphError> {
```
dkhofer, line 215: This long function could also use a doc comment

dkhofer:
What is the overall goal with these changes?  I'm partway through reviewing them and it seems to be to find all paths to a given GraphNodePoint from another one.  So that needs expansion and getting distances.  Is that approximately right?

gen-models/src/region.rs
dkhofer: :+1:

gen-models/src/region.rs
```diff
@@ -573,4 +653,580 @@ mod tests {
...
+                        phased: 0,
+                    },
+                ],
+            );
+
+            (conn, bg.id)
+        }
+
+        #[test]
+        fn within_node() {
```
dkhofer: This is a style thing, but it helps me if the names of the tests are prefixed with `test_`.  There are a bunch of helper methods above this one, I almost missed that this was a test

gen-models/src/region.rs
```diff
@@ -573,4 +653,580 @@ mod tests {
...
+                kind: ResolvedRegionKind::Accession,
+                anchor_start,
+                anchor_end,
+                feature_length,
+                start,
+                end,
+            }
+        }
+
+        /// Creates a branched graph: {AAA,GGG} → TTT → {CCC,ATC}
```
dkhofer, line 865: I think it would also be interesting to try a topology like AAA -> {CC, GGGG} -> TTT (with varying sequence lengths in the middle)

dkhofer (approved):
Overall LGTM, but needs more doc comments, plus a few minor style and other things.  Plus I think it'd be good to test against a branch -> rejoin type of graph topology

gen-graph/src/lib.rs
Chris7: I can call it segment  length. block start/end is the cumulative length

gen-graph/src/lib.rs
Chris7: good idea

PR #166, opened 2026-05-30
Agent guidance

Chris7:


bobvh:
should we gitignore .cursor and .kiro too then?

PR #167, opened 2026-06-01
Graph fix

Chris7:
There were a few cases where graph construction was failing. The root of it is `blocks_from_edges` assumes a pairwise set of edges, so if there were more source nodes than target nodes (2 upstream nodes going to the same point), it would fail.

### DIFF
--- a/gen-annotations/src/translate/bed.rs
+++ b/gen-annotations/src/translate/bed.rs
@@ -9,7 +9,7 @@ use gen_graph::{GraphNode, project_path};
 use gen_models::{
     block_group::BlockGroup,
     db::GraphConnection,
-    errors::PathError,
+    errors::{BlockGroupError, PathError},
     reference_alias::{ReferenceAlias, ReferenceAliasError},
     sample::Sample,
 };
@@ -29,6 +29,8 @@ pub enum BedError {
     ReferenceAliasError(#[from] ReferenceAliasError),
     #[error("Error loading path blocks: {0}")]
     PathError(#[from] PathError),
+    #[error("Error loading block group graph: {0}")]
+    BlockGroupError(#[from] BlockGroupError),
 }
 
 pub fn translate_bed<R, W>(
@@ -73,7 +75,7 @@ where
                 Entry::Occupied(entry) => entry.into_mut(),
                 Entry::Vacant(entry) => {
                     let path = BlockGroup::get_current_path(conn, &bg.id);
-                    let graph = BlockGroup::get_graph(conn, &bg.id);
+                    let graph = BlockGroup::get_graph(conn, &bg.id)?;
                     let mut tree = IntervalTree::default();
                     let mut position: i64 = 0;
                     for (node, strand) in project_path(&graph, &path.blocks(conn)?) {
--- a/gen-models/src/block_group.rs
+++ b/gen-models/src/block_group.rs
@@ -221,11 +221,7 @@ impl IntervalTreeSource for BlockGroupTreeSource {
         &self,
         conn: &GraphConnection,
     ) -> Result<IntervalTree<i64, NodeIntervalBlock>, BlockGroupError> {
-        Ok(BlockGroup::intervaltree_for(
-            conn,
-            &self.block_group_id,
-            self.remove_ambiguous_positions,
-        ))
+        BlockGroup::intervaltree_for(conn, &self.block_group_id, self.remove_ambiguous_positions)
     }
 }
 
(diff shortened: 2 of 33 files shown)


gen-models/src/edge.rs
```diff
@@ -310,48 +311,151 @@ impl Edge {
         }
     }
 
-    fn get_block_boundaries(
```
Chris7, line 313: moved to tests, it's only used there now

gen-models/src/block_group.rs
```diff
@@ -1880,6 +1880,175 @@ mod tests {
         assert!(BlockGroupEdge::edges_for_block_group(conn, &block_groups[0].id).is_empty());
     }
 
+    #[test]
+    fn test_get_graph_branched_graph() {
+        // Branched graph: {AAA,GGG} → TTT → {CCC,ATC}
```
Chris7, line 1884: the test case that started all this. I found that when building the blocks up, if start/ends were not evenly matched we would drop nodes.

gen-models/src/edge.rs
```diff
@@ -310,48 +311,151 @@ impl Edge {
...
-        source_edges: Option<&Vec<&Edge>>,
-        target_edges: Option<&Vec<&Edge>>,
-    ) -> Vec<i64> {
-        let mut block_boundary_coordinates = HashSet::new();
-        if let Some(actual_source_edges) = source_edges {
-            for source_edge in actual_source_edges {
-                block_boundary_coordinates.insert(source_edge.source_coordinate);
-            }
+    pub fn edges_for_block_group_nodes(
```
Chris7, line 319: new query function to lookup nodes for a blockgroup. I use this in here to expand a graph if the augmented edge ids are insufficient to build a complete graph. One though was to have this incomplete graph be explicit and add GraphNodes with a node_id belonging to something we mark as incomplete so we can have lazier expansion.

gen-annotations/src/translate/bed.rs
```diff
@@ -73,7 +75,7 @@ where
                 Entry::Occupied(entry) => entry.into_mut(),
                 Entry::Vacant(entry) => {
                     let path = BlockGroup::get_current_path(conn, &bg.id);
-                    let graph = BlockGroup::get_graph(conn, &bg.id);
+                    let graph = BlockGroup::get_graph(conn, &bg.id)?;
```
Chris7, line 78: bunch of stuff from changing this to a result

gen-models/src/edge.rs
```diff
@@ -310,48 +316,162 @@ impl Edge {
...
+            });
+        }
+        if coordinates.len() == 1 {
+            return Ok(vec![(coordinates[0], coordinates[0])]);
+        }
+
+        Ok(coordinates.into_iter().tuple_windows().collect())
+    }
+
+    pub fn blocks_from_edges(
```
Chris7, line 394: block_group_id was added to blocks_from_edges to lookup nodes if needed to finish the graph for lazy graph construction where we start with a set of edges that are partial positions on a graph, like an accession. It's a bit non-deterministic in that you can deliberately not pass in edges to have a deliberate partial graph. e.g. if you give it a bunch of edges for node foo, it assumes that's all the information it needs about node foo. But if you give a node that only ever appears as a source/exit node, it'll look it up so it can correctly construct it in the graph. If we feel strongly about this, we could alternatively have an `IncompleteGraphNode` type that then expands that part of the graph if it's used.

gen-annotations/src/translate/bed.rs
dkhofer: :+1:

gen-models/src/block_group.rs
dkhofer: :+1: Is the fact that we're not including the start/end nodes part of the problem?  All the code I wrote assumed those would be present

gen-models/src/edge.rs
dkhofer: :+1:

gen-models/src/edge.rs
dkhofer: :+1:

gen-models/src/edge.rs
dkhofer: :+1:

dkhofer: I'm good with keeping this change and seeing how it goes

dkhofer (approved):
LGTM.  This all seems reasonable, and if there are things about it we decide we don't like, it won't be too bad to change

PR #171, opened 2026-06-02
Preserve file suffix

Chris7:
Preserves file suffixes. This way we don't overwrite compression suffixes and the parser only records the type of file if it knows it.


gen-models/src/assets.rs
```diff
@@ -268,14 +268,12 @@ pub trait AssetUri {
         &self,
         workspace: &Workspace,
         checksum: &HashId,
-        file_type: FileTypes,
```
Chris7, line 271: file_type is derived when storing now and as needed from the uri.

gen-models/src/assets.rs
```diff
@@ -284,6 +281,27 @@ pub trait AssetUri {
         workspace: &Workspace,
     ) -> Result<(), FileStoreError>;
 
+    fn hashed_filename(&self, checksum: &HashId) -> String {
+        self.asset_filename(checksum)
+    }
+
+    fn suffix(&self) -> Option<String> {
```
Chris7, line 288: suffix is always just taken from the uri, not passed in.

gen-models/src/assets.rs
```diff
@@ -292,36 +310,27 @@ pub trait AssetUri {
...
-        Ok(workspace
-            .asset_dir()?
-            .join(Self::asset_filename(checksum, file_type)))
+    fn asset_filename(&self, checksum: &HashId) -> String {
+        let suffix = self.suffix().unwrap_or_default();
+        let suffix = suffix.strip_prefix('.').unwrap_or(&suffix);
+        if suffix.is_empty() {
+            return checksum.to_string();
+        }
+        format!("{checksum}.{suffix}")
```
Chris7, line 319: now we'll leave something like .fa.gz if it is the suffix

gen-models/src/file_types.rs
```diff
@@ -118,22 +118,36 @@ impl From<gen_models_capnp::FileType> for FileTypes {
...
-            Some("gb") | Some("gbk") | Some("genbank") => FileTypes::GenBank,
-            Some("fa") | Some("fasta") | Some("fna") => FileTypes::Fasta,
-            Some("gfa") => FileTypes::GFA,
-            Some("gaf") => FileTypes::GAF,
-            Some("vcf") => FileTypes::VCF,
-            Some("csv") => FileTypes::CSV,
-            Some("gff") | Some("gff3") => FileTypes::Gff3,
-            Some("bed") => FileTypes::Bed,
-            Some("tbi") => FileTypes::Tabix,
+        for extension in extensions.iter().rev() {
```
Chris7, line 130: this is now an iterator b/c it can be .fa.tar.gz, etc so we just look at things separated by `.` to find the extension

gen-models/src/assets.rs
dkhofer: :+1:

gen-models/src/assets.rs
dkhofer: :+1:

gen-models/src/assets.rs
dkhofer: :+1:

gen-models/src/file_types.rs
dkhofer: :+1:

gen-models/src/operations.rs
```diff
@@ -2830,7 +2870,15 @@ mod tests {
         let outside_checksum = calculate_file_checksum(outside_file.path()).unwrap();
         let outside_asset_path = format!(
             ".gen/assets/{outside_checksum}.{}",
-            FileTypes::suffix(FileTypes::Fasta)
+            outside_file
```
dkhofer: I feel like I'm seeing this pattern in a couple places for getting the suffix of a file.  Might be good to write a utility function for it

dkhofer (approved):
LGTM, one minor suggestion

gen-models/src/operations.rs
Chris7: actually can replace this w/ the one we already have. will do

PR #175, opened 2026-06-05
Accession edge refactor

Chris7:
Refactor accessions to point directly at the edge table instead of rercreating its own. Most of the downstream work is kind of routine. The exception is in the genbank annotation import/export. This works for now, but is a place I want to update to a central api for adding annotations. The current system for adding annotations is too limited and cannot support things like passing in metadata, multi-fragment features, and non-path based updates.


src/exports/genbank.rs
```diff
@@ -11,7 +11,7 @@ use std::{
 
 use gb_io::{self, seq::Location};
 use gen_annotations::projection::{
-    AnnotationSegment, accession_edges_to_segments, project_annotation_segments,
+    AnnotationSegment, accession_blocks_to_segments, project_annotation_segments,
```
Chris7, line 14: the annotations api is going to be updated in another PR. This gets the genbank import/export working with the new system. What is in here might be ported to be the new api. But some issues with the current annotation api is it's path based, doesn't do multi-segment features from the api side, doesn't capture format specific metadata.

gen-models/src/block_group.rs
Chris7, line 622: this function is going to be rewritten at some point w/ an improved accession creation interface

gen-models/src/accession.rs
```diff
@@ -781,6 +680,50 @@ mod tests {
...
+        assert_eq!(edges.len(), 5);
+        assert!(edges.iter().all(|edge| edge.accession_id == accession.id));
+        assert_eq!(
+            edges
+                .iter()
+                .map(|edge| edge.index_in_path)
+                .collect::<Vec<_>>(),
+            vec![0, 1, 2, 3, 4]
+        );
+        assert!(
```
dkhofer, line 721: Is this assert necessary?  Seems like it's just confirming the schema change went through

gen-models/src/block_group.rs
dkhofer: :+1:

src/exports/genbank.rs
dkhofer: :+1:

src/exports/genbank.rs
```diff
@@ -425,6 +457,10 @@ pub fn export_genbank(
             }
         }
 
+        if annotation_path_blocks.is_empty() {
```
dkhofer, line 460: Can you put a comment here?  Is the assumption that if we didn't find anything for an annotation, we put the annotation on the whole sequence?

gen-models/src/block_group.rs
```diff
@@ -650,63 +636,95 @@ impl BlockGroup {
...
+            .expect("should find path edge entering accession start block");
+        let (end_edge_index, end_edge) = all_path_edges
+            .iter()
+            .enumerate()
+            .find(|(_, edge)| {
+                edge.source_node_id == end_block.node_id
+                    && edge.source_coordinate == end_block.sequence_end
+            })
+            .expect("should find path edge leaving accession end block");
+        assert!(
```
dkhofer, line 658: I asked codex to review this branch, and it points out there are some places in non-test code, like here, where `assert!` is called.  It'd be good to replace them with returning an Error result

src/imports/genbank.rs
```diff
@@ -110,360 +77,244 @@ fn annotation_group_name(
...
+        .expect("should find path edge entering annotation start block");
+    let (end_edge_index, end_edge) = all_path_edges
+        .iter()
+        .enumerate()
+        .find(|(_, edge)| {
+            edge.source_node_id == end_block.node_id
+                && edge.source_coordinate == end_block.sequence_end
+        })
+        .expect("should find path edge leaving annotation end block");
+    assert!(
```
dkhofer, line 119: Another non-test assert

gen-models/src/accession.rs
```diff
@@ -340,18 +301,19 @@ impl Accession {
...
-            where ap.accession_id = ?1 \
-            order by ap.index_in_path;";
-        AccessionEdge::query(conn, query, params![accession_id])
+    pub fn get_edges_by_id(
+        conn: &GraphConnection,
+        accession_id: &HashId,
+    ) -> Result<Vec<AccessionEdge>, AccessionError> {
+        Ok(AccessionEdge::query(
+            conn,
+            "select * from accession_edges where accession_id = ?1",
```
dkhofer, line 310: Even though it doesn't look like it matters to the code that calls this method, I think it'd be good to order the sql results by `index_in_path`.  (Another catch by codex.)

dkhofer (approved):
Overall LGTM, just a few minor comments

PR #177, opened 2026-06-10
Remove layout dead space with a global shrink-only compression pass

bobvh:
Unedited output of Claude 5 Fable below. TL;DR: the combination of a combinatorial grid with a node that bypasses that grid causes a weird layout artefact. This fixes that by looking for empty bands, which means that it should also stop any other artefacts like this from showing up in the future.

> 
> Dense graphs (layers connected all-to-all) exposed two layout distortions: a large empty band between the grid and a bypass node, and inflated column spacing when a wide node shares a Sugiyama layer. Both stem from the Brandes-Köpf coordinate assignment: its four extreme layouts disagree on block placement for dense graphs (diagonal staircase blocks), and averaging them leaves slack that no individual layout had.
> 
> Add compress_dead_space, a global shrink-only pass that collects occupied bands from node bounding boxes per axis and shrinks gaps exceeding vertex_spacing down to vertex_spacing. It runs both before and after redistribute_horizontal_chains: before, because re-centered nodes straddle inter-column gaps and block compression; after, because redistribution can vacate inflated columns.
> 
> This pass complements rather than replaces the edge router's compress_graph, which is a label-aware normalizer that may expand spacing so routing channels clear wide nodes (verified empirically: removing it corrupts rendering). Cross-referencing comments document the distinction.
> 
> New snapshot tests cover all-to-all grids with a bypass node (4x4, 5x4, 4x5, and variable widths), plus a zoom round-trip test proving disperse/contract still scales spacing and restores the original layout. All existing snapshots are unchanged.


gen-tui/src/distribute_nodes.rs
```diff
@@ -667,6 +667,120 @@ pub fn redistribute_horizontal_chains(
...
+/// layer pair can see. Removing either reintroduces distortions: without
+/// `compress_graph`, junctions can end up inside wide nodes; without this
+/// pass, dense graphs keep large dead bands.
+///
+/// In the layout pipeline this runs both before and after
+/// [`redistribute_horizontal_chains`]: before, because once nodes are
+/// re-centered along their chains they straddle inter-column gaps and block
+/// compression; after, because redistribution can vacate bands (e.g. a wide
+/// node moving out of an inflated column), reopening dead space.
+pub fn compress_dead_space(
```
bobvh, line 718: This is the actual new code. I added extra comments here because otherwise it may seem like a redundant step.

PR #178, opened 2026-06-10
Fix bug where files weren't being manifested during pull

dkhofer:
This fixes a bug I found where eg "foo.fa" was in the original repo but wasn't showing up when I did a gen clone of that repo from genhub

### DIFF
--- a/gen-models/src/manifest.rs
+++ b/gen-models/src/manifest.rs
@@ -50,15 +50,6 @@ impl<'a> Capnp<'a> for ManifestOperationFileAddition {
 }
 
 impl ManifestOperationFileAddition {
-    fn from_legacy_file_addition(file_addition: FileAddition) -> Self {
-        let operation_file = OperationFile::new(file_addition.file_path());
-        Self {
-            file_addition,
-            filename: operation_file.filename,
-            file_path: operation_file.file_path,
-        }
-    }
-
     pub fn get_files_for_operation(
         conn: &OperationsConnection,
         operation_hash: &HashId,
@@ -66,15 +57,18 @@ impl ManifestOperationFileAddition {
         Ok(
             OperationFile::get_files_for_operation(conn, operation_hash)?
                 .into_iter()
-                .map(|file| Self {
-                    file_addition: FileAddition {
+                .map(|file| {
+                    let file_addition = FileAddition {
                         id: file.id,
                         asset_uri: file.asset_uri,
                         file_type: file.file_type,
                         checksum: file.checksum,
-                    },
-                    filename: file.filename,
-                    file_path: file.file_path,
+                    };
+                    Self {
+                        file_path: file_addition.file_path().to_string(),
+                        file_addition,
+                        filename: file.filename,
+                    }
                 })
                 .collect(),
         )
@@ -147,16 +141,6 @@ impl<'a> Capnp<'a> for ManifestOperation {
         let mut operation_builder = builder.reborrow().init_operation();
         self.operation.write_capnp(&mut operation_builder);
 
-        let mut legacy_file_additions_builder = builder
-            .reborrow()
-            .init_file_additions(self.file_additions.len() as u32);
-        for (i, file_addition) in self.file_additions.iter().enumerate() {
-            let mut file_addition_builder = legacy_file_additions_builder.reborrow().get(i as u32);
-            file_addition
-                .file_addition
-                .write_capnp(&mut file_addition_builder);
-        }
-
         let mut operation_file_additions_builder = builder
             .reborrow()
             .init_operation_file_additions(self.file_additions.len() as u32);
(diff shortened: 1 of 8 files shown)


src/commands/clone.rs
```diff
@@ -20,8 +20,12 @@ use crate::{
 const ORIGIN: &str = "origin";
 
 pub fn execute(url: &str) -> Result<(), Box<dyn std::error::Error>> {
+    execute_in_parent(url, Path::new("."))
+}
+
+fn execute_in_parent(url: &str, parent_path: &Path) -> Result<(), Box<dyn std::error::Error>> {
```
Chris7: Can the workspace from main.rs be passed through to give the current directory instead of doing parent_path. It's functionally the same but seems like it would tie into the stuff we're already reusing better.

Chris7: Meaning I think you can derive a new workspace with the workspace + repo_name, then ensure_gen_dir() handles the creation

src/commands/clone.rs
```diff
@@ -139,6 +151,66 @@ mod tests {
         assert!(repo_path.is_dir());
     }
 
+    #[test]
+    fn clone_from_file_remote_materializes_asset_files() {
```
Chris7: prefix with `test_`? :)

src/operation_management.rs
```diff
@@ -1334,6 +1274,62 @@ fn copy_operation_from_remote_fs(
...
+            operation_file_path,
+            stored_asset_path.to_string_lossy().to_string(),
+            user_dst_path.to_string_lossy().to_string(),
+        )
+    })?;
+
+    Ok(())
+}
+
+fn operation_file_destination(file_addition: &ManifestOperationFileAddition) -> String {
```
Chris7: i think it would be clearer to call this `operation_file_addition` instead of `file_addition`

src/operation_management.rs
```diff
@@ -2824,5 +2901,56 @@ mod tests {
...
+                asset_dir.join("abc123.fa")
+            );
+            assert_eq!(
+                remote_asset_destination(&asset_dir, ".gen/assets/abc123.fa").unwrap(),
+                asset_dir.join("abc123.fa")
+            );
+        }
+
+        #[test]
+        fn test_materialize_operation_file_uses_filename_for_legacy_asset_path() {
```
Chris7: we shouldn't support legacy stuff

src/operation_management.rs
```diff
@@ -1396,27 +1391,48 @@ fn download_remote_operation_assets(
...
+            materialize_operation_file(&workspace, file_addition, &stored_asset_path)?;
         }
     }
 
     Ok(())
 }
 
+fn remote_asset_destination(
+    asset_dir: &FilePath,
+    asset_path: &str,
```
Chris7: why don't we keep relative paths intact?

src/operation_management.rs
```diff
@@ -2824,5 +2901,56 @@ mod tests {
...
+            let context = setup_gen();
+            let asset_dir = context.workspace().asset_dir().unwrap();
+
+            assert_eq!(
+                remote_asset_destination(&asset_dir, "abc123.fa").unwrap(),
+                asset_dir.join("abc123.fa")
+            );
+            assert_eq!(
+                remote_asset_destination(&asset_dir, "assets/abc123.fa").unwrap(),
+                asset_dir.join("abc123.fa")
```
Chris7: not sure i understand this. Shouldn't we want to keep relative paths?

Chris7 (commented):
A few questions on path handling. One note -- this could result in a lot of stuff being dumped in the top level of the repository for places that have hundreds of operations of files that ended up in .assets only (e.g. import X from a path outside of repo)

gen-models/src/manifest.rs
```diff
@@ -50,31 +50,25 @@ impl<'a> Capnp<'a> for ManifestOperationFileAddition {
 }
 
 impl ManifestOperationFileAddition {
-    fn from_legacy_file_addition(file_addition: FileAddition) -> Self {
```
dkhofer, line 53: Most of the changes in this file are just cleanup after the capnp change above

gen-models/src/manifest.rs
```diff
@@ -740,6 +661,49 @@ mod tests {
         assert_eq!(manifest.operations[0].operation.hash, op1.hash);
     }
 
+    #[test]
```
dkhofer, line 664: I *think* this test would have passed in any case, but AFAICT there wasn't anything testing file addition path stuff here

src/graphs/translation.rs
```diff
@@ -1469,7 +1469,7 @@ fn translate_from(
 
 #[cfg(test)]
 mod tests {
-    use std::collections::{HashMap, HashSet};
+    use std::collections::HashSet;
```
dkhofer, line 1472: Just fixes a warning currently on main

src/operation_management.rs
```diff
@@ -572,39 +572,7 @@ fn apply_operations_to_remote(
                     )
                 })?;
 
-                // This is materializing the file in the directory. If file_addition file_path
```
dkhofer, line 575: AFAICT this is just pure refactoring.  There is a superficial rework of the logic checking that the file path is the right kind to copy to, but it doesn't change the result of the function

src/operation_management.rs
```diff
@@ -1395,28 +1374,55 @@ fn download_remote_operation_assets(
...
                 Some(auth_token),
-                &file.file_path,
+                stored_asset_path,
             )?;
         }
-        if !user_destination.exists() {
-            std::fs::copy(destination.as_path(), user_destination.as_path())?;
+    }
+
+    for file_addition in &manifest_operation.file_additions {
```
dkhofer: This is the actual bug fix

src/operation_management.rs
```diff
@@ -1395,28 +1374,53 @@ fn download_remote_operation_assets(
...
                 destination.as_path(),
                 Some(auth_token),
-                &file.file_path,
+                stored_asset_path,
             )?;
         }
-        if !user_destination.exists() {
-            std::fs::copy(destination.as_path(), user_destination.as_path())?;
+
+        let stored_asset_actual_path = repo_root.join(file_addition.file_addition.file_path());
```
dkhofer, line 1405: This is the actual bug fix

src/operation_management.rs
Chris7: can you keep the comment?

Chris7: or update it for the docstirng for the new func

Chris7 (approved):
lgtm. just one comment about a docstring

src/operation_management.rs
dkhofer: Yep, added as docstring to the new function

PR #179, opened 2026-06-11
Refactor annotation access: DB-backed Annotation on SequenceGraph

bobvh:
Affects Python and R bindings only.

- Consolidates annotations that came from the DB or files into a common type
- `list_annotations()` is now exposed on `SequenceGraph` in both Python and R, so annotations are reachable directly from the graph rather than only via the widget.

Introduction notebooks and vignettes are updated to use the new API.

This is the base for the stacked protein-translation PR.

### DIFF
--- a/gen-python/examples/annotations.ipynb
+++ b/gen-python/examples/annotations.ipynb
@@ -5,11 +5,7 @@
    "id": "intro",
    "metadata": {},
    "source": [
-    "# Annotation listing and navigation\n",
-    "\n",
-    "This notebook imports pUC19 from a GenBank file, loads its feature annotations,\n",
-    "and navigates to the MCS (multiple cloning site) using `widget.go_to()` and\n",
-    "`widget.show()`."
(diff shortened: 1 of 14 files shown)


gen-python/src/python_api/translation.rs
```diff
@@ -0,0 +1,257 @@
...
+    let parsed = Region::parse(region)
+        .map_err(|e| PyRuntimeError::new_err(format!("Invalid region '{region}': {e}")))?;
+
+    let bg_ids = ancestor_block_group_ids(conn, bg_id);
+    let mut all_matches: Vec<Annotation> = Vec::new();
+    let mut seen: HashSet<HashId> = HashSet::new();
+    for id in &bg_ids {
+        for ann in Annotation::query(
+            conn,
+            "SELECT a.* FROM annotations a \
```
bobvh: I'm looking into replacing/deleting this since it's a raw query.

bobvh: This was replaced

PR #180, opened 2026-06-11
Protein translation: core, CLI, Python & R surfaces

bobvh:
Adds translation of a sequence (sub)graph into a protein `SequenceGraph`, applying reading frame, strand, and NCBI codon tables.

Stacked on #179 — it depends on the new annotation coordinate fields and `Annotation::resolve`. Review/merge #179 first; the diff here is against that branch.

### Commits
- **Core** — junction-aware translation in `src/graphs/translation.rs`: `CodonTable`, `TranslationParams`, `translate_block_group`/`translate_annotation` (annotation-bounded sub-DAG extraction), and the shared `with_translation_operation` wrapper that records the work as a tracked database operation.
- **CLI** — a new `gen derive` command group housing the translation subcommand.
- **Python** — `SequenceGraph.translate_annotation()`, accepting a region string or an `Annotation` object.
- **R** — the matching `translate_annotation()` entry point in the R bindings.
- **Notebook** — a worked `protein_translation.ipynb` example.

### Breaking change
The top-level `gen derive-subgraph` and `gen derive-chunks` commands move under the new group: `gen derive subgraph` and `gen derive chunks`. Behaviour is unchanged; only the command path changes.


gen-python/src/python_api/block_group.rs
```diff
@@ -524,6 +531,184 @@ impl PySequenceGraph {
...
+    ///     Collection for the protein block group. Defaults to this graph's collection.
+    /// output_sample : str
+    ///     Sample name for the protein block group. Required.
+    /// strand : str, optional
+    ///     ``"forward"`` or ``"reverse"``. Inferred from the annotation when omitted.
+    /// frame : int
+    ///     Initial reading frame offset (0, 1, or 2). Default: 0.
+    /// codon_table : int
+    ///     NCBI codon table ID. Default: 1 (Standard).
+    #[pyo3(signature = (region=None, output_collection=None, output_sample=None, strand=None, frame=0, codon_table=1, start=None, end=None))]
```
bobvh: the bulk of this is translating between Python's default arguments and Rust

gen-r/src/rust/src/lib.rs
```diff
@@ -2285,6 +2352,160 @@ impl SequenceGraph {
...
+    /// @param end Exclusive end coordinate in path space. Defaults to path
+    ///   length when NULL. Must be <= path length. Default: NULL.
+    /// @param output_collection Collection for the protein block group.
+    ///   Defaults to this graph's collection.
+    /// @param output_sample Sample name for the protein block group. Required.
+    /// @param strand `"forward"` or `"reverse"`. NULL infers from the annotation.
+    /// @param frame Initial reading frame offset: 0, 1, or 2.
+    /// @param codon_table NCBI codon table ID (default: 1 = Standard).
+    /// @return A new SequenceGraph containing the protein sequence.
+    fn translate_annotation(
```
bobvh, line 2355: Like in Python: this is just to handle all the default=None arguments

src/commands/derive/chunks.rs
```diff
@@ -0,0 +1,56 @@
+use clap::Args;
+
+use crate::commands::{
+    cli_context::CliContext, graph_operations::derive_chunks::derive_chunks_operation,
+};
+
+/// Derive subgraphs at specified breakpoints
+#[derive(Debug, Args, Clone)]
+pub struct Command {
```
bobvh, line 9: signatures copied verbatim from CLI on main (is still missing a name-collection rename PR)

src/commands/graph_operations/derive_translation.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub collection: Option<String>,
+    pub sample: String,
+    pub region: String,
+    pub output_sample: String,
+    pub strand: Option<String>,
+    pub frame: u8,
+    pub codon_table: u8,
+}
+
+pub fn derive_translation_operation(
```
bobvh, line 23: I can't call this directly from Python/R because the interface is very different (not string based),  we do have the shared transaction/operation wrapper in with_translation_operation (src/graphs/translation.rs) that all 3 sources (python/r/cli)

src/graphs/translation.rs
```diff
@@ -0,0 +1,2364 @@
...
+    operations::OperationInfo,
+    path::revcomp,
+    sample::{NewSample, Sample},
+    sequence::Sequence,
+    session_operations::{end_operation, start_operation},
+};
+use petgraph::{graphmap::DiGraphMap, visit::EdgeRef};
+use thiserror::Error;
+
+// ── CodonTable ───────────────────────────────────────────────────────────────
```
bobvh: hardcoding these because they should cover 99.9% of cases

src/graphs/translation.rs
```diff
@@ -0,0 +1,2364 @@
...
+    let subgraph = extract_full_graph(conn, block_group_id)?;
+    let strand = params.strand.unwrap_or(Strand::Forward);
+    let label = HashId::convert_str(&format!("translate-full:{block_group_id}"));
+    let bg_name = format!("{}-protein", bg.name);
+    translate_core(conn, subgraph, strand, params, &bg_name, label)
+}
+
+/// Extract the entire block-group graph (everything between the start and end
+/// terminals) as an in-memory sub-DAG. No database writes.
+fn extract_full_graph(
```
bobvh, line 609: The translation takes place in graph space, on a subgraph that can be defined through a number of routes. For a full blockgroup that may look silly here, but it allows me to keep the same input format.

The comment refers to the fact that it's not a regular subgraph like from derive-subgraph since making those modifies the database.

src/graphs/translation.rs
```diff
@@ -0,0 +1,2364 @@
...
+        edge_chromosome_indices,
+        entry_nodes: entry_set.into_iter().collect(),
+        exit_nodes: exit_set.into_iter().collect(),
+    })
+}
+
+/// Extract the sub-DAG between an annotation's entry and exit coordinates from the
+/// block group's GenGraph, capturing every variant branch in between. No database
+/// writes. The entry/exit nodes are trimmed to the annotation's coordinates.
+fn extract_annotation(
```
bobvh: @dkhofer @Chris7 this is where I derive a subgraph from an annotation (so actually accession) like we talked about. It requires a GenGraph (no spidering). If we end up going further down the lazyloading path this would be a target to look at or share with derive-subgraph.

src/graphs/translation.rs
```diff
@@ -0,0 +1,2364 @@
...
+    #[error("Operation tracking failed: {0}")]
+    OperationTracking(#[from] OperationError),
+}
+
+/// Run a translation closure inside a DB transaction and operation record.
+///
+/// Rolls back both connections on any failure. This is the canonical wrapper
+/// used by the CLI, Python, and R bindings — call it instead of hand-rolling
+/// `start_operation` / `end_operation` around translation.
+pub fn with_translation_operation<F>(
```
bobvh, line 914: @dkhofer does this match your earlier request to keep transaction and operation management outside of the CLI/python/R?

src/graphs/translation.rs
```diff
@@ -0,0 +1,2364 @@
...
+    Ok(())
+}
+
+/// Walk the DNA sub-DAG forward from `from`, entering `node` at offset 0 with
+/// `pending` bases (0–2) already collected toward the current codon. Emits one
+/// junction node per straddling codon and connects to the next protein anchor (or
+/// `PATH_END`). Fully-consumed short nodes are skipped without producing nodes,
+/// which is what keeps the protein graph connected.
+#[expect(clippy::too_many_arguments, reason = "recursive graph walk state")]
+fn codon_walk(
```
bobvh, line 329: Recursive helper that creates the edges in the protein graph, it's not the main deal.

src/graphs/translation.rs
```diff
@@ -0,0 +1,2364 @@
...
+        .map_err(|e| TranslationOperationError::Transaction(e.to_string()))?;
+
+    Ok(protein_bg)
+}
+
+/// Shared translation core: fetch sequences, orient by strand, propagate reading
+/// frames, then build and persist the protein block group.
+///
+/// `label_hash` seeds the Merkle hash of nodes with no protein predecessor.
+fn translate_core(
```
bobvh: This is where the actual magic happens.

  1. Fetches sequences for every node (dna_by_node)
  2. Handles reverse-strand orientation (revcomp + flip edges)
  3. Propagates reading frames forward via Kahn's BFS (entry_frames)
  4. Computes per-(node, frame) codon runs (node_codons)
  5. Creates NodeIds with Merkle-chain hashing for synonymous-variant collapse (protein_node_ids)
  6. Wires up the protein graph's edges by calling codon_walk from each entry point and from each anchor node's trailing partial codon
  7. Dedupes and bulk-inserts the edges

src/graphs/translation.rs
```diff
@@ -0,0 +1,2364 @@
...
+            parent_block_group_id: None,
+            is_default: false,
+        },
+    )
+    .map_err(|e| TranslationError::BlockGroupError(e.to_string()))?;
+
+    let bg_id = protein_bg.id;
+    let mut bg_edges: Vec<BlockGroupEdgeData> = Vec::new();
+
+    // Merkle-chain hashing: each protein node's hash = hash(sorted_predecessor_protein_hashes + aa_sequence).
```
bobvh: @Chris7  this is an example of Node id hashes that use only topology and sequence (we talked about it a while ago as a concept)

src/graphs/translation.rs
```diff
@@ -0,0 +1,2364 @@
...
+            nucleotide_index(codon[1]),
+            nucleotide_index(codon[2]),
+        ) else {
+            return b'X';
+        };
+        self.table[i1 * 16 + i2 * 4 + i3]
+    }
+}
+
+// ── Error ─────────────────────────────────────────────────────────────────────
```
Chris7: it's redundant to have a separator thats the same name as the struct below and is the only member of that divide. We don't really do this in any of the code

src/graphs/translation.rs
```diff
@@ -0,0 +1,2364 @@
...
+        self
+    }
+}
+
+// ── Internal helpers ──────────────────────────────────────────────────────────
+
+/// The complete codons a DNA node contributes at a given entry frame, plus the
+/// trailing bases that spill into successors.
+///
+/// `head_skip` leading bases finish a codon begun in a predecessor (handled by an
```
Chris7: don't undestand this comment. Is this phase

bobvh:
I just remembered that I need to doublecheck the NCBI codon tables. The format and their weird quaternary index checks out, but I haven't done a base for base confirmation that they're identical. 

Edit: one of them was wrong, fixed and added reference + test, sorry about not catching it immediately

gen-python/src/python_api/block_group.rs
dkhofer: :+1:

gen-python/src/python_api/block_group.rs
```diff
@@ -524,6 +531,184 @@ impl PySequenceGraph {
...
+                    })
+                    .map_err(|e| PyRuntimeError::new_err(e.to_string()))?
+                } else {
+                    with_translation_operation(ctx, &label, || {
+                        translate_block_group(conn, &bg_id, params)
+                    })
+                    .map_err(|e| PyRuntimeError::new_err(e.to_string()))?
+                }
+            }
+            Some(any) => {
```
dkhofer: I think it'd be more descriptive to name this variable "region" instead of "any"

gen-r/src/rust/src/lib.rs
```diff
@@ -1985,6 +2031,27 @@ struct SequenceGraph {
 unsafe impl Send for SequenceGraph {}
 unsafe impl Sync for SequenceGraph {}
 
+fn resolve_start_end(
```
dkhofer: This is basically the same as the version with the python bindings.  Maybe it could be refactored into gen-core somewhere?  Maybe lib.rs

gen-r/src/rust/src/lib.rs
dkhofer: :+1:

src/commands/derive/chunks.rs
dkhofer: :+1:

src/commands/graph_operations/derive_translation.rs
dkhofer: :+1:

src/graphs/translation.rs
dkhofer: :+1:

src/graphs/translation.rs
dkhofer: 👍

src/graphs/translation.rs
```diff
@@ -0,0 +1,2393 @@
...
+    for succ in &successors {
+        codon_walk(conn, bg_id, ctx, &junction_from, rem, *succ, ci, bg_edges)?;
+    }
+    Ok(())
+}
+
+fn make_edge(
+    conn: &GraphConnection,
+    bg_id: &HashId,
+    src: HashId,
```
dkhofer: Could you use "source" and "target" instead of "src" and "tgt" in the parameter names?  We tend to prefer full words

src/graphs/translation.rs
```diff
@@ -0,0 +1,2393 @@
...
+/// branches at the first and last columns are all represented.
+struct TranslationSubgraph {
+    graph: DiGraphMap<HashId, ()>,
+    node_ranges: HashMap<HashId, (HashId, Option<i64>, Option<i64>)>,
+    edge_chromosome_indices: HashMap<(HashId, HashId), Vec<i64>>,
+    entry_nodes: Vec<HashId>,
+    exit_nodes: Vec<HashId>,
+}
+
+fn is_terminal(node_id: HashId) -> bool {
```
dkhofer: You could just import the standard one: https://github.com/genhub-bio/gen/blob/main/gen-core/src/lib.rs#L38

src/graphs/translation.rs
dkhofer: :+1:

src/graphs/translation.rs
```diff
@@ -0,0 +1,2393 @@
...
+    let dna = ctx
+        .dna_by_node
+        .get(&node)
+        .map(|s| s.as_bytes())
+        .unwrap_or(&[]);
+    let successors: Vec<HashId> = ctx.subgraph.neighbors(node).collect();
+
+    if pending.is_empty() {
+        // At a codon boundary entering `node`: frame 0 for this node on this path.
+        if let Some(&pid) = ctx.protein_node_ids.get(&(node, 0)) {
```
dkhofer: Can you call this `protein_id` instead of `pid`?  As a software engineer, I have a very strong association with "pid" meaning "process id"

src/graphs/translation.rs
```diff
@@ -0,0 +1,2393 @@
...
+    )?);
+
+    let junction_from = WalkFrom {
+        id: junction_id,
+        coord: 1,
+        merkle: junction_merkle,
+    };
+    // After the codon, `node`'s entry frame on this path equals the bytes carried in.
+    let anchor_frame = pending.len() as u8;
+    if let Some(&pid) = ctx.protein_node_ids.get(&(node, anchor_frame)) {
```
dkhofer: Same request to rename pid as protein_id here

src/graphs/translation.rs
dkhofer: 👍

src/graphs/translation.rs
```diff
@@ -0,0 +1,2393 @@
...
+    conn: &GraphConnection,
+    annotation: &Annotation,
+    block_group_id: Option<&HashId>,
+    params: TranslationParams<'_>,
+) -> Result<BlockGroup, TranslationError> {
+    let bg_id = block_group_id.ok_or_else(|| {
+        TranslationError::BlockGroupError("translation requires a block group id".into())
+    })?;
+
+    let accession_edges = Accession::get_edges_by_id(conn, &annotation.accession_id);
```
dkhofer: I think this will need to change with Chris's latest accession work

src/graphs/translation.rs
dkhofer: Yeah, this is reasonable

src/graphs/translation.rs
dkhofer: 👍

dkhofer: Could you add that as a comment in the code?

src/graphs/translation.rs
```diff
@@ -0,0 +1,2393 @@
...
+) -> Result<BlockGroup, TranslationError> {
+    let TranslationSubgraph {
+        graph: subgraph,
+        node_ranges,
+        edge_chromosome_indices,
+        entry_nodes,
+        exit_nodes,
+    } = subgraph;
+
+    // Collect real node IDs needed for sequence fetch.
```
dkhofer, line 1006: This is part 1?  Please add that to this comment

src/graphs/translation.rs
```diff
@@ -0,0 +1,2393 @@
...
+        let seq = sequences_by_node
+            .get(real_nid)
+            .ok_or_else(|| TranslationError::Sequence(format!("missing node {real_nid}")))?;
+        let slice = seq
+            .get_sequence(*seq_start, *seq_end)
+            .map_err(|e| TranslationError::Sequence(e.to_string()))?;
+        dna_by_node.insert(vid, slice);
+    }
+
+    // Reverse strand: rev-comp sequences and flip graph edges; swap entry/exit.
```
dkhofer: This is part 2?

src/graphs/translation.rs
```diff
@@ -0,0 +1,2393 @@
...
+                .chunks_exact(3)
+                .map(|c| params.codon_table.translate_codon(c))
+                .collect();
+            let tail = body[split..].to_vec();
+            node_codons.insert((node_id, frame), NodeCodons { aa, tail });
+        }
+    }
+
+    // Step 5 & 6: create protein BlockGroup and persist nodes/edges
+    Sample::get_or_create(
```
dkhofer: We have sample lineage now: https://github.com/genhub-bio/gen/blob/main/gen-models/src/sample_lineage.rs

Should this new sample be a child of the one with the DNA sequence graph?

bobvh: Good point, Iet's get Chris' take. I think so?

src/graphs/translation.rs
```diff
@@ -0,0 +1,2393 @@
...
+    /// empty block group of the given name.
+    fn new_block_group(name: &str) -> (GraphConnection, BlockGroup) {
+        let conn = get_connection(None).expect("should open in-memory database");
+        Collection::create(&conn, "test").unwrap();
+        let bg = create_bg(&conn, "test", Sample::DEFAULT_NAME, name);
+        (conn, bg)
+    }
+
+    /// Save a DNA sequence as a node. Returns the node id and its length.
+    fn mk_node(conn: &GraphConnection, dna: &str, tag: &str) -> (HashId, i64) {
```
dkhofer: please rename to `make_node` or `build_node`.  Terse method names just make it harder to understand the code.  Same with `mk_edge` and `bge` below

src/graphs/translation.rs
```diff
@@ -0,0 +1,2393 @@
...
+        let graph = BlockGroup::get_graph(conn, bg_id).expect("should load protein graph");
+        let start = graph.nodes().find(|n| n.node_id == PATH_START_NODE_ID);
+        let end = graph.nodes().find(|n| n.node_id == PATH_END_NODE_ID);
+        match (start, end) {
+            (Some(start), Some(end)) => has_path_connecting(&graph, start, end, None),
+            _ => false,
+        }
+    }
+
+    /// Every full PATH_START → PATH_END amino-acid string, read left to right.
```
dkhofer: How is this different from BlockGroup's get_all_sequences()?

src/graphs/translation.rs
bobvh: Locally I've rebased onto 185 to prepare for that.

dkhofer (approved):
Overall looks good, but I had some comments you should address (or explain why it's best not to).  I know you're aware that the accession code is changing, so please keep coordinating with Chris on when to rebase and update to use the new methods

src/graphs/translation.rs
bobvh: Correction: Sample lineage is used to resolve annotations, not for biological interpretation.

gen-core/src/lib.rs
```diff
@@ -47,6 +47,28 @@ pub fn is_end_node(node_id: HashId) -> bool {
     node_id == PATH_END_NODE_ID
 }
 
+/// Clamp `start`/`end` to a path of `path_length`, defaulting `start` to 0 and
+/// `end` to `path_length` when omitted. Returns an error message on out-of-range
+/// or inverted bounds; callers wrap it in their own error type.
+pub fn resolve_start_end(
```
Chris7: this kind of validation here is a bit confusing. We have places that do a variant of this, the Region. Why this?

src/graphs/translation.rs
```diff
@@ -0,0 +1,2394 @@
...
+struct WalkCtx<'a> {
+    subgraph: &'a DiGraphMap<HashId, ()>,
+    dna_by_node: &'a HashMap<HashId, String>,
+    protein_node_ids: &'a HashMap<(HashId, u8), HashId>,
+    codon_table: &'a CodonTable,
+}
+
+/// Wire `from` to `PATH_END` unless it is `PATH_START` (a gene too short to encode
+/// a single residue contributes no edge).
+fn connect_to_end(
```
Chris7, line 302: doesn't really seem useful as a standalone function

src/graphs/translation.rs
```diff
@@ -0,0 +1,2394 @@
...
+        .map_err(|e| TranslationError::BlockGroupError(e.to_string()))?;
+
+    let entry_node_id = segments[0].node_id;
+    let entry_coord = segments[0].range.start;
+    let last_seg = segments.last().unwrap();
+    let exit_node_id = last_seg.node_id;
+    let exit_coord = last_seg.range.end;
+
+    // GraphNode slices that contain the annotation's entry/exit coordinates.
+    let entry_gn = gen_graph
```
Chris7: i think yoiu can get multiple entries here

src/graphs/translation.rs
```diff
@@ -0,0 +1,2394 @@
...
+    let bg_name = format!("{}-protein", annotation.name);
+    translate_core(conn, subgraph, strand, params, &bg_name, annotation.id)
+}
+
+/// Translate a coordinate range on a block group's current path into a protein
+/// sequence graph, in memory. The path's interval tree maps the path-space
+/// `start`/`end` coordinates to node-level entry/exit points; the same
+/// `extract_annotation` + `translate_core` pipeline used by `translate_annotation`
+/// is then applied to the resulting subgraph.
+pub fn translate_path_range(
```
Chris7: the translate stuff is pretty similar, might be better suited as an impl thing where you define the thing to translate as whatever gives me an intervaltree.

src/graphs/translation.rs
Chris7: is there a reason to have this be persistent? It's a derived operation

src/graphs/translation.rs
Chris7: its meant to track biology. that's why we use it for annotation inheritance

Chris7: Having this be a new sample is very odd. It should be a new blockgroup with protein sequence types (or honestly maybe the type on blockgroup). Or in the same blockgroup but a type field in block group edges

src/graphs/translation.rs
bobvh: Mainly protein engineering work, recoding to a different organism, annotating protein domains, ... For a good part of the field the underlying DNA sequence is irrelevant and they only care about the protein.

src/graphs/translation.rs
bobvh: done

gen-r/src/rust/src/lib.rs
bobvh: Done

src/graphs/translation.rs
bobvh: Done and added to the agent's memory with emphasis because it keeps forgetting

src/graphs/translation.rs
bobvh: Done

src/graphs/translation.rs
bobvh: Done, together with all the other shorthand lengthening.

src/graphs/translation.rs
bobvh: done

gen-core/src/lib.rs
bobvh: Because we're looking inside a specific block group, whereas the region resolves across multiple bg. That being said, I've moved this into Range where it makes more sense.

src/graphs/translation.rs
bobvh: Simplified this by ending walks at stop codons instead of defined end, and just picking the start codon itself as starting point, regardless of possible variants, since it's always M anyway.

src/graphs/translation.rs
bobvh: GPT-ism, fixed in agents.md too now

src/graphs/translation.rs
bobvh: fixed

src/graphs/translation.rs
bobvh: Not as pub no, but here its just so I don't copy paste the same 8 lines in 3 different spots.

src/graphs/translation.rs
bobvh: It doesn't do the pruning etc, but BlockGroup's method is the right choice, switched.

gen-annotations/src/projection.rs
```diff
@@ -40,6 +44,17 @@ impl From<&AccessionNode> for AnnotationSegment {
     }
 }
 
+/// Compute the annotation segments from accession nodes.
```
Chris7, line 47: It makes it from an annotation, not nodes

gen-core/src/range.rs
```diff
@@ -9,6 +9,24 @@ pub struct Range {
 }
 
 impl Range {
+    /// Build a `Range` from optional `start`/`end`, defaulting `start` to 0 and
+    /// `end` to `length` when omitted. Returns an error message on out-of-range
+    /// or inverted bounds; callers wrap it in their own error type.
+    pub fn clamped(start: Option<i64>, end: Option<i64>, length: i64) -> Result<Range, String> {
```
Chris7, line 15: Should use a RangeError

PR #182, opened 2026-06-11
gen-models: list annotations across a block group's sample lineage

bobvh:
TL;DR: adds a method to get all annotations associated with a block group and its ancestors (sample-based). It's based on the Resolve method, which I can't use because the question is reversed (I want to ask : give me all associated annotations given a block group, not "give me a block group that has this annotation")

Claude:

> Adds `Annotation::list_in_block_group_lineage(conn, collection_name, sample_name, block_group_name)` to gen-models: a single recursive-CTE method returning every annotation on a block group, walking the sample lineage (the block group plus its same-named ancestors), deduplicated by id and ordered closest-first.
> 
> This is the shared building block for the binding-layer de-DIY in #179 — the Python and R bindings call this instead of hand-rolled raw SQL in the binding layer. Scoped to match the existing listing behavior (collection + block-group name + sample lineage), modeled on `Annotation::resolve`'s CTE minus the name filter.
> 
> Tests cover self-listing, ancestor inheritance, block-group-name scoping, and exclusion of samples outside the lineage.

### DIFF
--- a/gen-models/src/annotations.rs
+++ b/gen-models/src/annotations.rs
@@ -379,6 +379,60 @@ impl Annotation {
         Ok(Annotation::query(conn, query, params![group]))
     }
 
+    /// List every annotation on a block group, following the sample lineage.
+    ///
+    /// Collects annotations attached to the block group named `block_group_name`
+    /// in `collection_name`, considering `sample_name` together with all of its
+    /// ancestor samples. This mirrors how a block group inherits its parents'
+    /// annotations. Results are de-duplicated by annotation id and ordered from
+    /// the closest sample in the lineage outward.
+    ///
+    /// # Arguments
+    ///
+    /// * `collection_name` - Collection the block group belongs to.
+    /// * `sample_name` - Sample whose lineage is walked, including the sample
+    ///   itself.
+    /// * `block_group_name` - Name shared by the block group across the lineage.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`AnnotationError::DatabaseError`] if the query fails.
+    pub fn list_in_block_group_lineage(
+        conn: &GraphConnection,
+        collection_name: &str,
+        sample_name: &str,
+        block_group_name: &str,
+    ) -> Result<Vec<Annotation>, AnnotationError> {
+        let query = "WITH RECURSIVE visible_samples(name, depth, path) AS (
+                 SELECT ?2, 0, ',' || ?2 || ','
+                 UNION ALL
+                 SELECT sl.parent_sample_name,
+                        visible.depth + 1,
+                        visible.path || sl.parent_sample_name || ','
+                 FROM sample_lineage sl
+                 JOIN visible_samples visible ON sl.child_sample_name = visible.name
+                 WHERE instr(visible.path, ',' || sl.parent_sample_name || ',') = 0
+             )
+             SELECT a.id,
+                    a.name,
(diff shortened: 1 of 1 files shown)


gen-models/src/annotations.rs
```diff
@@ -379,6 +379,60 @@ impl Annotation {
...
+    /// # Errors
+    ///
+    /// Returns [`AnnotationError::DatabaseError`] if the query fails.
+    pub fn list_in_block_group_lineage(
+        conn: &GraphConnection,
+        collection_name: &str,
+        sample_name: &str,
+        block_group_name: &str,
+    ) -> Result<Vec<Annotation>, AnnotationError> {
+        let query = "WITH RECURSIVE visible_samples(name, depth, path) AS (
```
bobvh, line 410: warrants extra attention, I'm not very familiar with recursive SQL queries.

gen-models/src/annotations.rs
```diff
@@ -379,6 +379,60 @@ impl Annotation {
...
+    ///
+    /// * `collection_name` - Collection the block group belongs to.
+    /// * `sample_name` - Sample whose lineage is walked, including the sample
+    ///   itself.
+    /// * `block_group_name` - Name shared by the block group across the lineage.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`AnnotationError::DatabaseError`] if the query fails.
+    pub fn list_in_block_group_lineage(
```
Chris7: I'd like this pattern if its possible:
* make this consistent with the other calls, using `query` instead of `list`. Name something like `query_with_lineage`
* Have the resolver function use that as well with the annotation name from region.name

gen-models/src/annotations.rs
```diff
@@ -1118,6 +1172,161 @@ mod tests {
...
+                .collect::<Vec<_>>();
+            BlockGroupEdge::bulk_create(conn, &block_group_edges);
+            let path = Path::create(conn, accession_name, &block_group.id, &edge_ids).unwrap();
+            let accession =
+                BlockGroup::add_accession(conn, &path, accession_name, 0, 5, cache).unwrap();
+            Annotation::get_or_create(conn, annotation_name, "genes", &accession.id, None).unwrap()
+        }
+
+        #[test]
+        fn lists_annotations_on_block_group() {
```
Chris7: can you prefix these with `test_`

gen-models/src/annotations.rs
```diff
@@ -1118,6 +1172,161 @@ mod tests {
         }
     }
 
+    mod lineage_listing {
```
Chris7, line 1179: what about the behavior with 2 annotations with the same name that exists in both parent and child.

Chris7 (approved):
I think you can reuse the query across the resolve as well.

gen-models/src/annotations.rs
bobvh: Leave it up to the user to handle. I don't think we want to do versioning of annotations by spreading them out over samples for example.

bobvh: Test added to show what happens in case we want to change it.

gen-models/src/annotations.rs
bobvh: I did the rename, but I don't think reusing it in the resolver is possible because they're querying different things. There is some repeated code now, but I can't isolate a whole query. I added comments to clarify that for future reference.

PR #183, opened 2026-06-11
Reuse nodes when deriving a subgraph from a block group

dkhofer:


src/graphs/operators.rs
```diff
@@ -308,6 +303,48 @@ pub fn make_stitch(
 ) -> Result<(), GraphOperationError> {
     let conn = context.graph().conn();
 
+    let stitch_inputs = stitch_inputs(conn, collection_name, parent_sample_name, region_names)?;
+    validate_stitch_inputs(&stitch_inputs)?;
+    let block_group_chunks = stitch_inputs
+        .iter()
+        .map(|input| input.chunk.clone())
```
dkhofer, line 310: In this context, a chunk is a block group that's going to be stitched together with other chunks (block groups).  If it helps I can add a comment to that effect

src/graphs/operators.rs
```diff
@@ -318,6 +355,72 @@ pub fn make_stitch(
...
+            )));
+        }
+    }
+
+    Ok(stitch_inputs)
+}
+
+/// Given a list of block groups to stitch together, confirms there are no
+/// duplicate block groups (which would generate a cycle in the graph) or edges
+/// that are shared between block groups
```
Chris7, line 385: is there a reason not to support overlapping edges? Wouldn't it just be a merge operation then?

src/graphs/operators.rs
```diff
@@ -881,4 +986,155 @@ mod tests {
         // path3 + path2 concatenated
         assert_eq!(path5.sequence(conn).unwrap(), "ATCGATCAAGGAACACATCAATCG");
     }
+
+    #[test]
+    fn make_stitch_rejects_duplicate_region_input() {
```
Chris7: test_ prefix

Chris7 (approved):
Can we have a test that verifies there's no new nodes after?

src/graphs/operators.rs
dkhofer: If sequence graph A has an edge from node 1 at bp 123 to node 2 at bp 456 and that same edge is in sequence graph B, and we stitch A to B, then there's a way to get from B back to A (follow the edge to node 2 at bp 456) and we have a cycle.  The general check for having a cycle would also catch this, but it seems like an obvious thing to check and warn about

dkhofer:
_Can we have a test that verifies there's no new nodes after?_

Around line 3976 of block_group.rs, there's a check that the number of nodes is the same before and after calling derive_subgraph.  That's the method that the commands call.  Was there somewhere else you wanted me to add a check?

src/graphs/operators.rs
dkhofer: :tableflip:

Fixed

PR #185, opened 2026-06-15
Accession api on accession nodes

Chris7:
An attempt to make a better api for creating accessions. There's a way to make an accession directly from a Region as well as an intervaltree now.

### DIFF
--- a/gen-core/src/range.rs
+++ b/gen-core/src/range.rs
@@ -2,7 +2,7 @@ use std::cmp::{max, min};
 
 use itertools::Itertools;
 
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub struct Range {
     pub start: i64,
     pub end: i64,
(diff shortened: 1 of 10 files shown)


gen-core/src/range.rs
```diff
@@ -2,7 +2,7 @@ use std::cmp::{max, min};
 
 use itertools::Itertools;
 
-#[derive(Clone, Debug, Eq, Hash, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
```
Chris7, line 5: copy trait added, so a few places will have clone dropped

bobvh:
While getting my branches ready I ran into the bindings breaking and 2 small bugs (something that fails doc tests, and a panic that wasn't propagated upwards). Patch attached: 
[accession-node-api-combined-fixes.patch](https://github.com/user-attachments/files/29104813/accession-node-api-combined-fixes.patch)

gen-core/src/range.rs
dkhofer: :+1:

dkhofer (approved):
This is all pretty straightforward, LGTM.  The build needs some fixes though

PR #186, opened 2026-06-15
Use nodes for accessions instead of edges

Chris7:
Replaces #175 

Refactors the accession model to store data on nodes instead of using an edge table. This makes accessions an ordered array of node slices. The strand information is stored on the accession node, as this was formerly on the edges. I think at some point we may have to do some more strand work to track it when we convert a region/accession into an intervaltree

### DIFF
--- a/gen-annotations/src/projection.rs
+++ b/gen-annotations/src/projection.rs
@@ -27,19 +27,17 @@ impl OrderedMerge for AnnotationSegment {
     }
 }
 
-pub fn accession_nodes_to_segments(nodes: &[AccessionNode]) -> Vec<AnnotationSegment> {
-    nodes
-        .iter()
-        .filter(|node| node.sequence_end > node.sequence_start)
-        .map(|node| AnnotationSegment {
+impl From<&AccessionNode> for AnnotationSegment {
+    fn from(node: &AccessionNode) -> Self {
+        AnnotationSegment {
             node_id: node.node_id,
             range: Range {
                 start: node.sequence_start,
                 end: node.sequence_end,
             },
             strand: node.strand,
-        })
-        .collect()
+        }
+    }
 }
 
 pub fn project_annotation_segments(
--- a/gen-annotations/src/projection.rs
+++ b/gen-annotations/src/projection.rs
@@ -5,7 +5,7 @@ use gen_core::{
     path::PathBlock,
     range::{OrderedMerge, Range, merge_ordered_items},
 };
-use gen_models::accession::AccessionEdge;
+use gen_models::accession::AccessionNode;
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct AnnotationSegment {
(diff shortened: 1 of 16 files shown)


gen-annotations/src/projection.rs
```diff
@@ -27,50 +27,17 @@ impl OrderedMerge for AnnotationSegment {
...
-                        end: segment_end,
-                    },
-                    strand,
-                });
-            }
-        }
-
-        if edge.target_coordinate < 0 {
-            break;
+impl From<&AccessionNode> for AnnotationSegment {
```
Chris7, line 30: I kept the AnnotationSegment for now even though it is very similar to the AccessionNode.

gen-models/src/accession.rs
```diff
@@ -431,7 +401,13 @@ impl RegionResolver for Accession {
 
         match matches.len() {
             0 => Err(RegionResolutionError::NotFound(region.name.clone())),
-            1 => Ok(matches.into_iter().next().unwrap()),
```
Chris7, line 434: removed the unwrap

src/updates/vcf.rs
```diff
@@ -715,7 +715,7 @@ mod tests {
             false,
         )
         .unwrap();
-        update_with_vcf(
+        let _operation = update_with_vcf(
```
Chris7, line 718: clippy stuff for whatever reason

src/updates/vcf.rs
```diff
@@ -1302,7 +1301,13 @@ mod tests {
             vec![Sample::DEFAULT_NAME.to_string()],
             false,
         )
-        .unwrap();
+        .unwrap_err();
```
Chris7, line 1304: update b/c of making somethign a result

src/updates/vcf.rs
```diff
@@ -1245,8 +1245,7 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "Unable to create accession")]
```
Chris7, line 1248: we test the err now instead of a panic

gen-annotations/src/projection.rs
dkhofer: :+1:

gen-models/src/accession.rs
```diff
@@ -455,180 +431,115 @@ impl Query for Accession {
     }
 }
 
-impl AccessionEdge {
+impl AccessionNode {
```
dkhofer, line 466: Since, as we established offline, it's not clear to non-bio people like me who are familiar with the data model but not the need for an accession to span all exons in a coding sequence, could you add a comment giving that context here, for why we're using nodes and not edges to define an accession?

dkhofer: (Assuming there isn't a better place to put a comment like that.)

src/updates/vcf.rs
dkhofer: :+1:

src/updates/vcf.rs
dkhofer: :+1:

src/updates/vcf.rs
dkhofer: :+1:

dkhofer (approved):
Overall LGTM, apart from one or two comments.  It took me a while to understand the context but I'm fine with this

PR #188, opened 2026-06-22
fix(transaction): rollback if tx_commit fails in run_write and __exit__

bobvh:
## Summary
- A commit failure (e.g. constraint violation at flush time) left the SQLite connection mid-transaction, so the next write failed with "cannot start a transaction within a transaction" instead of a clear error.
- `run_write` and `Repository.__exit__` now roll back before propagating the commit error.


dkhofer (approved):
LGTM!

PR #192, opened 2026-06-23
Vcf optimizations

Chris7:
Optimize vcf updates to reduce lookups

### DIFF
--- a/src/updates/vcf.rs
+++ b/src/updates/vcf.rs
@@ -42,6 +42,8 @@ use crate::{
     progress_bar::{add_saving_operation_bar, get_handler, get_progress_bar},
 };
 
+const VCF_CHANGE_APPLY_CHUNK_SIZE: usize = 5_000;
+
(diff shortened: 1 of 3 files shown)


src/updates/vcf.rs
```diff
@@ -162,15 +164,32 @@ fn prepare_change(
...
         preserve_edge,
+    }
+}
+
+fn lookup_path_region(
+    path_region_cache: &mut HashMap<Path, ResolvedGenRegion>,
+    conn: &GraphConnection,
+    sample_path: &Path,
+) -> Result<ResolvedGenRegion, BlockGroupError> {
+    Ok(match path_region_cache.entry(sample_path.clone()) {
```
dkhofer: I think this method doesn't need to return a Result given tho Ok is the only return value?

dkhofer (approved):
One minor thing, otherwise LGTM

PR #199, opened 2026-06-24
Add profiling

Chris7:
Adds a profiling command to identify slow points in the codebase. Adds instrumentation for some functions so they can be profiled. There are 2 profile modes: instrumentation which tracks things with the cfg profile attribute, and a sample based profiler.

profiler is enabled by building with `--features profiling`

### DIFF
--- a/bin/benchmark.sh
+++ b/bin/benchmark.sh
@@ -1,10 +1,16 @@
-#!/bin/sh
+#!/usr/bin/env bash
 set -e
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 BASE_DIR="$(dirname "${SCRIPT_DIR}")"
 GEN_DIR="${BASE_DIR}/.gen"
 GEN_BIN="${BASE_DIR}/target/release/gen"
+PROFILE_BIN="${BASE_DIR}/target/debug/gen"
+PYTHON_BIN="${PYTHON_BIN:-$(command -v python3 || command -v python)}"
+RESULTS=""
+PROFILE_DIR="${BASE_DIR}/.benchmark-profiles"
+PROFILE_RESULTS=""
+PROFILE_BENCHMARKS="${PROFILE_BENCHMARKS:-1}"
 
 init_test() {
   cd ${BASE_DIR}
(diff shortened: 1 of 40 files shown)


bin/benchmark.sh
```diff
@@ -13,51 +19,250 @@ init_test() {
 }
 
 get_size () {
-  local filesize_mb=$(python -c "import os;size=os.path.getsize('${GEN_DIR}/default.db') / (1024 * 1024);print(f'{size:.4f}')")
-  echo $filesize_mb
+  GEN_DIR="${GEN_DIR}" GEN_DB_PATH="${GEN_DIR}/gen.db" DEFAULT_DB_PATH="${GEN_DIR}/default.db" ${PYTHON_BIN} -c "import os
```
dkhofer, line 22: Is it possible to do all this with system commands, instead of invoking python?

bin/benchmark.sh
```diff
@@ -13,51 +19,250 @@ init_test() {
...
+for root, _, files in os.walk(gen_dir):
+    for name in files:
+        total += os.path.getsize(os.path.join(root, name))
+
+print(f'{total / (1024 * 1024):.4f} {size_mb(gen_db_path)} {size_mb(default_db_path)}')"
 }
 
 time_taken() {
-  local start_time=$(python -c "import time;print(round(time.time()*1000))")
+  local start_time=$(${PYTHON_BIN} -c "import time;print(round(time.time()*1000))")
```
dkhofer, line 44: Looks like the system command `date +%s%N` gives time in nanoseconds?

bin/benchmark.sh
```diff
@@ -13,51 +19,250 @@ init_test() {
...
-printf "%-35s %-10s %-10s\n" "Full import" ${FULL_IMPORT} ${FULL_SIZE}
-printf "%-35s %-10s %-10s\n" "Shallow import" ${SHALLOW_IMPORT} ${SHALLOW_SIZE}
-printf "%-35s %-10s %-10s\n" "HG00096 Update" ${HG96_IMPORT} ${HG96_SIZE}
-printf "%-35s %-10s %-10s\n" "HG00097 Update" ${HG97_IMPORT} ${HG97_SIZE}
-printf "%-35s %-10s %-10s\n" "HG00096 + HG00097 Update" ${SUM} ${BOTH_SIZE}
\ No newline at end of file
+read -r BOTH_SIZE BOTH_GEN_DB_SIZE BOTH_DEFAULT_DB_SIZE <<< "$(get_size)"
+record_result "HG00096 + HG00097 Update" "${SUM}" "${BOTH_SIZE}" "${BOTH_GEN_DB_SIZE}" "${BOTH_DEFAULT_DB_SIZE}"
+
+# echo "switch branch benchmark"
```
dkhofer, line 244: Why are these commented out?  Is the idea to uncomment to benchmark specific ones?

src/profiling.rs
```diff
@@ -0,0 +1,332 @@
...
+        entry.total += duration;
+    }
+
+    fn print_report(&self) {
+        let stats = self.stats.lock().unwrap_or_else(|err| err.into_inner());
+        let mut rows = stats
+            .iter()
+            .map(|(stack, stat)| (stack_frames(stack), stat.calls, stat.total))
+            .collect::<Vec<_>>();
+        rows.sort_by(|left, right| right.2.cmp(&left.2).then_with(|| left.0.cmp(&right.0)));
```
dkhofer, line 70: It'd be good to add a comment saying what fields 0 and 2 are

src/profiling.rs
```diff
@@ -0,0 +1,332 @@
...
+
+struct ProfilerLayer {
+    profiler: Profiler,
+}
+
+impl<S> Layer<S> for ProfilerLayer
+where
+    S: Subscriber + for<'span> LookupSpan<'span>,
+{
+    fn on_new_span(&self, attrs: &Attributes<'_>, id: &Id, ctx: Context<'_, S>) {
```
dkhofer: Style nitpick: ctx -> context

src/profiling.rs
```diff
@@ -0,0 +1,332 @@
...
+    calls: u64,
+    total: Duration,
+}
+
+struct SpanState {
+    stack: String,
+    starts: Vec<Instant>,
+}
+
+impl Profiler {
```
dkhofer, line 44: Is there a comment somewhere saying what the difference is between the Profiler and the SampleProfiler?  Maybe add a docstring to each class?

dkhofer (approved):
Overall LGTM, just a few comments

bin/benchmark.sh
Chris7: it's not natively supported in osx

bin/benchmark.sh
Chris7: i think it's this way just for portability of code between systems

PR #201, opened 2026-06-25
Fix pull problems

dkhofer:
This improves file download a bit, by refactoring the logic around the downloading into a separate method, and then using the new method to download annotation asset files to the assets directory.  I'm not doing annotation file manifesting here, it will take a schema change and maybe some more logic changes.

The new code also adds error handling for a couple cases

### DIFF
--- a/src/operation_management.rs
+++ b/src/operation_management.rs
@@ -1497,6 +1497,8 @@ fn download_remote_file_addition_asset(
     Ok(Some(destination))
 }
 
+// TODO: Add a parameter to force overwrite the local with the remote, in case
+// the user gets stuck and wants to get the remote copy
 fn verify_downloaded_asset(
     file_addition: &FileAddition,
     destination: &FilePath,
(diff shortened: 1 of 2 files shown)


src/operation_management.rs
```diff
@@ -1411,47 +1419,108 @@ fn download_remote_operation_assets(
...
+        destination.as_path(),
+        Some(auth_token),
+        stored_asset_path,
+    )?;
+    verify_downloaded_asset(file_addition, &destination)?;
+
+    Ok(Some(destination))
+}
+
+fn verify_downloaded_asset(
```
Chris7, line 1502: should add a todo to have a force here to overwrite local w/ remote, otherwise they will eternally be stuck

Chris7 (approved):
lgtm

src/operation_management.rs
dkhofer: Added

PR #205, opened 2026-07-01
Overhaul README and fix vignette style

bobvh:
## Summary

- Rewrites the top-level README around a vhs-recorded terminal demo, replacing the graph-theory framing; updates gen-python/README.md and gen-r/README.md to match.
- Adds a macos-15-intel release target so the README no longer needs to caveat lack of Intel Mac support.
- Points the CLI install instructions at the nightly release assets directly, matching genhub.bio's installation page, and clarifies that Windows users should run the Linux binary under WSL.
- Strips em dashes from the introduction and yeast_expression_library R vignettes.


README.md
```diff
@@ -1,120 +1,131 @@
...
-proteins in a proteome, or pieces of DNA in a cloning reaction mixture. Molecules generally start out as a node that
-holds the reference sequence, and new edges and nodes are added for every sequence variant that is designed or observed.
-To reconstitute a linear sequence, the client walks from node to node along a defined _path_. This data model allows the
-representation of a wide range of biological complexity -- from a single molecule to complex cultivars and cell lines,
-including any naturally occurring variation in addition to intended engineering.
 
-![Figure 1](docs/figures/figure_1.svg)
+Gen brings version control to genetic sequences. You can clone repositories, create branches, make edits, and push changes to a shared remote using the same workflow developers know from Git. It works across FASTA files, VCFs, GenBank records, and other common bioinformatics formats. Under the hood, Gen stores data as a sequence graph, allowing a single repository to represent a reference genome, known variants, and engineered modifications without repeatedly storing the same sequence.
```
dkhofer: I think more could be said, as early as possible, about what the user can do with gen.  One suggestion would be to change the opening to "Gen brings version control to genetic sequences.  With it, you can track variants and edits to whole genomes across multiple generations and lineages."  Maybe also instead of "You can clone repositories", "You can clone genome repositories".  Just to make it totally clear that the repos are for genomes or related sequences

README.md
```diff
@@ -1,120 +1,131 @@
...
+
+# Browse the graph in the terminal
+gen view
+
+# Push to a remote GenHub repository
+gen push
 ```
-pip install gen
+
+Python and R libraries expose the same operations:
```
dkhofer: Could you add a sentence like "For instance, here is some python code that uses gen:".  Just to make it clear that although we mentioned python and R, this is python specifically

README.md
dkhofer: Could you also add something about suitability for AI use?  Something like "Gen is suitable for AI or automation workflows that need structured history of sequence edits."  (That was an AI suggestion, feel free to use it or not.)

dkhofer:
I had a couple comments, but overall this is looking a lot more like I would expect a readme to!  Thanks!

README.md
bobvh: Done. I also made it one of the screenshots.

bobvh: Good suggestions. Implemented

README.md
bobvh: done

PR #207, opened 2026-07-06
ci: fix triggers for stacked PRs, add tests for python bindings, fix build for R bindings, add windows tests

bobvh:
- test.yml and r-package-check.yml now trigger on pull_request for any branch (branches: ["**"]) instead of only PRs targeting main, matching test-notebooks.yml and test-python.yml.
- test-python.yml adds windows-latest to its test matrix, with a per-OS capnp install step and fail-fast: false.
- publish-r-packages-tag.yaml:
  - adds a macos-15-intel / macos-x86_64 build target alongside the existing macos-arm64 one;
  - installs anyhtmlwidget from GitHub before building (a Remotes: dependency that R CMD INSTALL does not resolve), which was failing every platform;
  - puts x86_64-pc-windows-gnu Rust target onto the Windows job, Rtools' mingw toolchain needs it
  - gives each uploaded package a unique per-platform artifact name

### DIFF
--- a/.github/workflows/publish-pypi-tag.yaml
+++ b/.github/workflows/publish-pypi-tag.yaml
@@ -40,7 +40,7 @@ jobs:
 
       - name: Locate wheel build run
         id: wheels-run
-        uses: actions/github-script@v9
+        uses: actions/github-script@v7
         env:
           WHEEL_SHA: ${{ github.sha }}
           WHEEL_TAG: ${{ env.PUBLISH_TAG }}
@@ -70,7 +70,7 @@ jobs:
             core.setOutput("run_id", run.id.toString());
 
       - name: Download wheel artifacts
-        uses: actions/download-artifact@v8
+        uses: actions/download-artifact@v4
         with:
           pattern: wheels-*
           path: wheelhouse
@@ -84,7 +84,7 @@ jobs:
           ls -la dist
 
       - name: Set up Python
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@v5
         with:
           python-version: "3.x"
 
--- a/.github/workflows/publish-pypi-tag.yaml
+++ b/.github/workflows/publish-pypi-tag.yaml
@@ -40,7 +40,7 @@ jobs:
 
       - name: Locate wheel build run
         id: wheels-run
-        uses: actions/github-script@v7
+        uses: actions/github-script@v9
         env:
           WHEEL_SHA: ${{ github.sha }}
           WHEEL_TAG: ${{ env.PUBLISH_TAG }}
@@ -70,7 +70,7 @@ jobs:
             core.setOutput("run_id", run.id.toString());
 
       - name: Download wheel artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v8
         with:
           pattern: wheels-*
           path: wheelhouse
@@ -84,7 +84,7 @@ jobs:
           ls -la dist
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: "3.x"
 
--- a/.github/workflows/publish-r-packages-tag.yaml
+++ b/.github/workflows/publish-r-packages-tag.yaml
@@ -59,7 +59,7 @@ jobs:
 
       - name: Install R package dependencies
         env:
-          GITHUB_PAT: ${{ github.token }}
+          GITHUB_PAT: ${{ secrets.MINIMAL_RO_PAT }}
         run: |
           Rscript -e "install.packages(c('remotes', 'testthat'), repos = 'https://cloud.r-project.org')"
           Rscript -e "remotes::install_github('keller-mark/anyhtmlwidget@be23d31587b3b0986febef00ba27567e46dc3338', upgrade = 'never')"
--- a/.github/workflows/publish-r-packages-tag.yaml
+++ b/.github/workflows/publish-r-packages-tag.yaml
@@ -42,7 +42,7 @@ jobs:
           fi
 
       - name: Checkout ref
-        uses: actions/checkout@v7
+        uses: actions/checkout@v4
         with:
           ref: ${{ env.PACKAGE_REF }}
           fetch-depth: 0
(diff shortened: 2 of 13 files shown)


bobvh:
### Verification of the R publish workflow fix
To see that the R publishing workflow works now: https://github.com/genhub-bio/gen/actions/runs/28820186676

- All three legs green: macOS arm64, macOS x86_64, Windows. They all install the built package and run `testthat`
- Assets uploaded to the v0.2.0 release and all download with HTTP 200:
  - `genr_0.2.0-macos-arm64.tgz`
  - `genr_0.2.0-macos-x86_64.tgz`
  - `genr-windows-0.2.0.zip`

The Linux build is tested in docker-install-test as it was before

.github/workflows/test.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches: [ "main" ]
   pull_request:
-    branches: [ "main" ]
+    branches: [ "**" ]
```
bobvh, line 7: So checks get triggered even if your PR targets a different branch.

gen-python/src/python_api/repository/mod.rs
```diff
@@ -373,7 +373,10 @@ mod python_tests {
         pyo3::prepare_freethreaded_python();
         Python::with_gil(|py| {
             let tmp_dir = tempdir().unwrap();
-            let path = tmp_dir.path().to_str().unwrap().to_string();
+            // Escape backslashes so a Windows path (e.g. `C:\Users\...`) survives
```
bobvh, line 376: only actual change needed for windows

.github/workflows/publish-r-packages-tag.yaml
```diff
@@ -77,9 +81,9 @@ jobs:
       - name: Upload macOS binary artifact
         uses: actions/upload-artifact@v4
         with:
+          name: genr-${{ matrix.asset_suffix }}
           path: ${{ env.ASSET_NAME }}
           if-no-files-found: error
-          archive: false
```
Chris7, line 82: why is archive reenabled?

.github/workflows/test-notebooks.yml
```diff
@@ -11,22 +11,39 @@ env:
 
 jobs:
   test-notebooks:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
```
Chris7: keep this true

.github/workflows/test-python.yml
```diff
@@ -13,9 +13,9 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
-      fail-fast: true
```
Chris7, line 16: why are these changed

Chris7 (commented):
what's the reason for the tag along changes

bobvh:
> what's the reason for the tag along changes

Multiple CI/CD issues cropped up while editing the documentation, I collected them into one PR.

.github/workflows/publish-r-packages-tag.yaml
bobvh: It wasn't active. The action was at version V4, where that option just doesn't exist: https://github.com/actions/upload-artifact/blob/v4/action.yml . It does exist in the later versions, so I took the liberty of bringing all actions to the latest version and put the archive lines back. Only the pypi deployment actions haven't been tested, but I expect we'll deploy the next couple of days?

.github/workflows/test-notebooks.yml
bobvh: I thought we wanted it false so we could see which parts failed, flipping it back to true everywhere.

.github/workflows/test-python.yml
bobvh: I thought seeing which platform failed if it's not all of them was more useful, but I reverted it now

.github/workflows/publish-r-packages-tag.yaml
```diff
@@ -50,15 +52,17 @@ jobs:
...
-          target: x86_64-pc-windows-gnu
 
       - name: Install system dependencies
         run: |
           brew install capnp
 
       - name: Install R package dependencies
-        run: Rscript -e "install.packages('testthat', repos = 'https://cloud.r-project.org')"
+        env:
+          GITHUB_PAT: ${{ github.token }}
```
bobvh: This is added because of the install_github() call to get the anywidget dependency. Github imposes stricter rate limits unless you authenticate. The official R-lib example workflows do it too: https://github.com/r-lib/actions/blob/v2-branch/examples/README.md 

Security-wise: if I understand it correctly, the risk is that the R-scripts exfiltrate the token during the brief window it is valid. I pinned the dependency version so we reduce the exposure to a supply chain attack there.

bobvh: Reference on it being short-lived: https://docs.github.com/en/actions/concepts/security/github_token

.github/workflows/publish-pypi-tag.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
 
       - name: Locate wheel build run
         id: wheels-run
-        uses: actions/github-script@v7
+        uses: actions/github-script@v9
```
Chris7: why are these bumped now?

gen-r/Dockerfile
```diff
@@ -33,7 +34,8 @@ COPY gen-capnp-schemas ./gen-capnp-schemas
 COPY gen-r ./gen-r
 
 RUN Rscript -e "install.packages(c('testthat', 'knitr', 'rmarkdown'), repos = 'https://cloud.r-project.org')"
-RUN Rscript -e "if (!requireNamespace('remotes', quietly=TRUE)) install.packages('remotes', repos='https://cloud.r-project.org'); remotes::install_github('keller-mark/anyhtmlwidget', upgrade='never')"
+RUN --mount=type=secret,id=github_pat \
```
Chris7: how does this behave locally now?

Chris7 (commented):
can you confirm how the dockerfile behaves locally now with the secret addition to its local mount and add a reason to bump all the other actions

gen-r/Dockerfile
bobvh: set up to fallback to unauthenticated now

.github/workflows/publish-pypi-tag.yaml
bobvh: unbumped everything per discussion now

PR #208, opened 2026-07-07
Fix reverse-strand annotation translation for genes embedded in larger nodes

bobvh:
There was a bug in translating a gene to protein when the gene annotation is on the reverse strand and does not run all the way to the end of a node (i.e. having its reversed start codon start at the rightmost end of the node). This fixes that by properly making the reverse workflow the mirror image of the forward process.

### DIFF
--- a/src/graphs/translation.rs
+++ b/src/graphs/translation.rs
@@ -809,13 +809,7 @@ fn extract_from_entry(
 /// `PATH_START`, and the anchor node is trimmed on its right so no sequence to
 /// the right of the anchor is read. `translate_from` then reverse-complements and
 /// flips the subgraph, so translation begins at the anchor (the start codon) and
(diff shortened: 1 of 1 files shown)


dkhofer:
This is fine, but an alternate solution would be to get the graph in memory, then reverse it (including the sequences referred to by nodes), and then you could run the same code for the forward strand on the reversed graph.  That may be a bit more involved, but I feel like it could be a useful general solution to handling reverse strand things.

If you need to get this merged urgently, I can approve, but I'm curious how well reversing the graph would work

bobvh:
Hmm.  We do have the graph in memory already, so it wouldn't complicate things too much to load it as rev comp instead. We'd need to be real sure that we don't introduce off by one errors at the region parsing stage, but that's testable. I'll keep this open for now.

PR #210, opened 2026-07-08
Materialize annotation files

dkhofer:
Ensure that annotation and annotation index files are materialized on pull/clone


gen-models/src/annotations.rs
```diff
@@ -868,17 +896,42 @@ impl AnnotationFile {
             input.file_type,
             input.checksum_override,
         )?;
-        let index_file_addition = input
-            .index_file_path
-            .as_deref()
-            .map(|path| FileAddition::get_or_create(workspace, conn, path, FileTypes::Tabix, None))
-            .transpose()?;
+        let index_file_addition = if let Some(index_file_path) = input.index_file_path.as_ref() {
```
dkhofer, line 899: I'm not sure where the map/transpose pattern came from?  If it's something preferred to the `if let Some(...) ...` pattern, I could switch over, but I find the latter to be easier to understand

gen-capnp-schemas/gen-models.capnp
```diff
@@ -235,6 +235,11 @@ struct ManifestAnnotationFileAddition {
     none @3 :Void;
     some @4 :FileAddition;
   }
+  filePath @5 :Text;
```
Chris7: we should have `filename` too to match the other operationfileaddition areas

Chris7 (approved):
lgtm. just 1 comment about whether we need filename too

gen-capnp-schemas/gen-models.capnp
dkhofer: Added

Issue #211, opened 2026-07-08
Gen update with sequence bug


bobvh:
Deleting a gene using the region resolver (`gen update sequence <empty> --region-name <name>`) works in the test cases, but fails silently when deleting a (sub)sequence from a combinatorial library. For example deleting the first base of a genetic part creates an edge that results in two split fragments that are wired together, but no bypass edge that encodes the deletion. The command exits 0 and prints "Updated with sequence." with no indication anything went wrong.

To reproduce from the repo root:

```
GEN=target/release/gen   # cargo build --release --bin gen first
REPO=$(git rev-parse --show-toplevel)
WORK=/tmp/gen-bug-repro && rm -rf "$WORK" && mkdir -p "$WORK" && cd "$WORK"

"$GEN" init
"$GEN" import fasta "$REPO/fixtures/simple.fa" --sample wt
"$GEN" add-annotation --name SITE --sample wt m123:7-20
"$GEN" update library --sample wt --new-sample design --region-name SITE \
    --library "$REPO/fixtures/combinatorial_design.csv" --parts "$REPO/fixtures/parts.fa"

# cds1/cds2/cds3 each now have 3 incoming edges (one from each of p1/p2/p3) — confirm with:
"$GEN" export gfa --sample design design.gfa

# Delete the first base of cds1's start codon
"$GEN" update sequence "" --sample design --new-sample deleted --region-name cds1:0-1
"$GEN" export gfa --sample deleted deleted.gfa
```

Deleted.gfa should contain a bypass edge from each of p1/p2/p3 directly to the larger fragment of cds1 (skipping the deleted base), alongside the original full-length route. It doesn't have that bypass edge.

Claude suggested possible cause:

> resolve_update_region/target_update_region in src/updates.rs set region.path = None for non-Path-kind regions, and ResolvedGenRegion::intervaltree (gen-models/src/region.rs ~L479-489) builds the tree from Accession::intervaltree, which is scoped to just the accession's own span (gen-models/src/accession.rs ~L488-497) — it has no knowledge of the wider block group graph. The deletion branch of BlockGroup::insert_change/set_up_new_edges (gen-models/src/block_group.rs ~L831-984) needs real predecessor/successor anchors outside the edited span to wire its bypass edge; when the accession's node is reached via edges from other nodes rather than being a sub-range of one larger node, that context isn't in the local tree, so the bypass edge is never created. Insertions don't hit this because they only need to wire into the two local split fragments, which the local tree does contain.

[...]

PR #215, opened 2026-07-10
Add reproduction test for issue #211 bypass-edge bug

bobvh:
## Summary
- Adds a Rust test that mirrors the bash reproduction from #211: import fasta, add an annotation, run `update library`, then delete the first base of one library part (`cds1:0-1`).
- Asserts (via `BlockGroup::get_all_sequences`) that a bypass-deletion variant exists for every entry point into the part (p1, p2, p3), matching what `gen export gfa` + human inspection should show.
- The test currently fails: only 10 of the expected 13 sequences are present, confirming the missing bypass edges described in #211.

Relates to #211 (reproduction only; root cause not fixed here).

## Test plan
- [x] cargo test test_deletion_inside_combinatorial_library_part_creates_bypass_for_every_entry_point (currently fails, demonstrating the bug)

### DIFF
--- a/gen-models/src/region.rs
+++ b/gen-models/src/region.rs
@@ -566,7 +566,9 @@ impl ResolvedGenRegion {
         tree: Option<&IntervalTree<i64, NodeIntervalBlock>>,
     ) -> Result<Vec<AugmentedEdgeData>, BlockGroupError> {
         match self.kind {
-            ResolvedRegionKind::Path | ResolvedRegionKind::BlockGroup => {
+            ResolvedRegionKind::Path | ResolvedRegionKind::BlockGroup
+                if self.start_anchors.is_none() || self.end_anchors.is_none() =>
+            {
                 let local_tree;
                 let tree = match tree {
                     Some(tree) => tree,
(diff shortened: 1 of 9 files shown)


dkhofer:
I significantly widened the scope of this PR because most update operations were susceptible to this bug.  The solution is to ensure we make a copy of any existing incoming or outgoing edge that goes to (or comes from) the new part of the node being exposed as part of the edit.

For fasta.rs and sequence.rs, I refactored pretty much all the code into shared functions in src/updates.rs.  There is some new logic that gets used by each of them in BlockGroup::insert_change, which calls plan_edges in region.rs, which figures out the appropriate new edges to add, and has the fix, which takes care of those operations.

For gaf.rs and gfa.rs, they don't call BlockGroup::insert_change, so each one gets a custom new method similar to Region::plan_edges to figure out if additional edges need to be added.

Some fixes were also needed in library.rs and vcf.rs.

The only operation which didn't have the issue was genbank.rs.

I encouraged codex to refactor code and have the operations use the refactored methods as much as possible, partly to ensure the bug fixes get propagated, partly because we haven't really audited this code to see how much is duplicated.  Let me know if I went too far anywhere.

dkhofer:
Closing because https://github.com/genhub-bio/gen/pull/232 resolved the issue

PR #216, opened 2026-07-14
Doltlite backend

Chris7:
A massive rewrite to use doltlite as the backend instead of sqlite. It also comes with native git-like capabilities that are wired in.


.cargo/config.toml
```diff
@@ -0,0 +1,3 @@
+[target.'cfg(target_os = "linux")']
```
Chris7, line 1: for linux builds, improves build time by using mold linker. There's lld or somtehing similar on osx if you want to play with that

.github/workflows/publish-crates-io.yaml
```diff
@@ -14,6 +14,9 @@ on:
 jobs:
   publish:
     runs-on: ubuntu-latest
+    env:
```
Chris7, line 17: sccache is setup in a few places, it caches compilation artifacts to improve build times

bin/benchmark.sh
```diff
@@ -241,28 +241,28 @@ SUM=$(echo "${HG96_IMPORT} + ${HG97_IMPORT}" | bc)
 read -r BOTH_SIZE BOTH_GEN_DB_SIZE BOTH_DEFAULT_DB_SIZE <<< "$(get_size)"
 record_result "HG00096 + HG00097 Update" "${SUM}" "${BOTH_SIZE}" "${BOTH_GEN_DB_SIZE}" "${BOTH_DEFAULT_DB_SIZE}"
 
-# echo "switch branch benchmark"
```
Chris7, line 244: these tests are no longer broken w/ doltlite being used

bin/benchmark_sequential_vcf_changes.sh
```diff
@@ -0,0 +1,131 @@
+#!/usr/bin/env bash
```
Chris7, line 1: benchmark of multiple files to look at build time changes in many sequential changes

docs/commands.md
```diff
@@ -2,49 +2,98 @@
 
 This controls default choices for the `gen` command line
 
-- database
```
Chris7, line 5: no more database flag

gen-annotations/src/translate/bed.rs
```diff
@@ -37,6 +37,7 @@ pub fn translate_bed<R, W>(
     conn: &GraphConnection,
     collection: &str,
     sample: &str,
+    history_ref: Option<&str>,
```
Chris7, line 40: history_ref is now present on most query fields. It provides the results as they existed at the provided ref

gen-annotations/Cargo.toml
```diff
@@ -16,5 +16,5 @@ noodles = { version = "0.101.0", features = ["bed", "core", "gff"] }
 thiserror = "2.0.18"
 
 [dev-dependencies]
-rusqlite = { version = "0.32.1", features = ["array", "bundled"] }
+rusqlite = { package = "rusqdoltlite", version = "0.40.6", features = ["array", "bundled", "fallible_uint"] }
```
Chris7: the swap out of rusqlite w/ rusqdoltlite

gen-core/src/config.rs
```diff
@@ -3,10 +3,10 @@ use std::{
     path::{Path, PathBuf},
 };
 
-use crate::{HashId, errors::ConfigError};
+use crate::errors::ConfigError;
 
-pub const CHANGESET_DIR_NAME: &str = "changesets";
```
Chris7, line 8: changesets are no longer a thing, everything is stored in the graph db

gen-core/src/lib.rs
```diff
@@ -47,6 +48,104 @@ pub fn is_end_node(node_id: HashId) -> bool {
     node_id == PATH_END_NODE_ID
 }
 
+#[derive(
+    Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, serde::Deserialize, serde::Serialize,
+)]
+pub struct CommitRef(pub String);
```
Chris7, line 66: these are types around the dolt strings for hashes/refs/etc.

gen-diff/src/graph.rs
```diff
@@ -1,28 +1,78 @@
 use std::collections::{HashMap, HashSet};
```
Chris7, line 92: i do not love the diff setup. It needs to be reimplemented now that changesets are gone. The outputs look correct so I'm letting this be for now but I will probably do a follow up PR proving to myself we can't do it simpler. How exactly we want to show this information is changing so I'm not going to obsess over it.

gen-diff/src/operations.rs
```diff
@@ -1,796 +1,1974 @@
 use std::collections::{HashMap, HashSet};
```
Chris7, line 60: also don't love this file and the organization of nodes/etc. with operations. I'm wonering if gen-diff should be merged to gen-models now that it's not a pure graph based diff. TBD

gen-models/migrations/config/01-initial/up.sql
```diff
@@ -1,61 +1,10 @@
 CREATE TABLE defaults (
```
Chris7, line 1: this effectively becomes a config db, which is why the connection was renamed to ConfigConnection

src/views/diff.rs
```diff
@@ -1,9 +1,13 @@
-use std::{collections::HashMap, io, time::Instant};
```
Chris7, line 1: This view was changed a bit -- I have 3 top level segments for samples: added, removed, and modified. Block groups are then placed under the sample entry. This way things like adding a new child sample w/o any changes to the block groups is visible in the diff.

src/views/diff_graph.rs
```diff
@@ -1,8 +1,8 @@
 use std::collections::HashSet;
```
Chris7, line 1: i record the status of a node/edge (added/removed/changed) in the node and edge itself now, so this changed a little bit and is hopefully has less interfence going on

src/lib.rs
```diff
@@ -40,96 +40,46 @@ pub use gen_graph as graph;
...
-    files::GenDatabase,
+    db::{ConfigConnection, GraphConnection},
+    history::dolt::{active_branch, checkout},
     migrations::{run_migrations, run_operation_migrations},
 };
 use noodles::vcf::variant::record::samples::series::value::genotype::Phasing;
-use rusqlite::{Connection, OptionalExtension};
+use rusqlite::{Connection, OpenFlags};
 
 pub fn get_connection(
```
Chris7, line 50: this new indirection exists to indicate pick the default branch (main)

.cargo/config.toml
dkhofer: :+1:

.github/workflows/publish-crates-io.yaml
dkhofer: :+1:

bin/benchmark.sh
dkhofer: Excellent

bin/benchmark_sequential_vcf_changes.sh
dkhofer: :+1:

docs/commands.md
dkhofer: :+1:

docs/commands.md
```diff
@@ -2,49 +2,113 @@
...
 
-Checkouts allow a user to migrate the database to different states. To move the database to a given operation, the
-command in `gen --db db_name.db checkout -b branch_name operation_id`. If no branch name is specified, the current
-branch will be used. The operation_id corresponds to the operation in `operations`.
\ No newline at end of file
+Checkouts allow a user to migrate the database to different states. To move the database to a given commit, use
+`gen checkout -b branch_name commit_id`. If no branch name is specified, the current branch will be used.
+The `commit_id` corresponds to a commit listed by `gen operations`.
+
+# Clone
```
dkhofer, line 68: Thanks for adding this and the rest here

gen-annotations/src/translate/bed.rs
dkhofer: :+1:

gen-core/src/config.rs
dkhofer: :+1: Great

gen-core/src/lib.rs
dkhofer: :+1:

gen-diff/src/graph.rs
dkhofer: :+1:

gen-diff/src/graph.rs
```diff
@@ -61,237 +132,975 @@ impl From<DiffGraphEdge> for GraphEdge {
...
+        BlockGroup::get_by_id(graph_conn, &block_group_id, Some(inputs.source_ref)).ok();
+    if source_block_group.is_none() && target_block_group.is_none() {
+        return Ok(None);
     }
-    for edge in changes.edges.iter() {
-        edges_by_id.insert(edge.id, (edge, true));
+    // Added or removed child block groups only exist on one endpoint. The
+    // opposite endpoint is represented by the child's parent so the diff
+    // compares the child graph against the graph it diverged from.
+    let source_parent_block_group_id = target_block_group
```
dkhofer, line 318: Why is this being called the source parent block group ID but being set to that of the target block group's parent ID

dkhofer: Is this because of what the comment is saying about endpoints?

dkhofer: It would help a lot if you could add a top-level comment for this method saying what an endpoint is and giving an overall description of what this method is doing.  The mid-level comments are OK but harder to read when there's no context for terminology

gen-diff/src/operations.rs
```diff
@@ -1,796 +1,2066 @@
...
-pub struct DbDiff {
-    pub db_path: String,
-    pub added_block_groups: Vec<BlockGroupDiff>,
-    pub removed_block_groups: Vec<BlockGroupDiff>,
+/// Selects the revision range semantics used to build a diff.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum DiffRange {
+    /// Compare the source and target endpoint states (`source..target`).
+    TwoDot,
+    /// Compare the source/target merge base with the target (`source...target`).
```
dkhofer, line 124: I'm unfamiliar with this syntax, if there's a description of it online could you put a link in a code comment?  If note, what is a "merge base"?

gen-diff/src/graph.rs
```diff
@@ -61,237 +132,975 @@ impl From<DiffGraphEdge> for GraphEdge {
...
+
+fn span_node_ids(spans: &HashSet<GraphNode>) -> HashSet<HashId> {
+    spans
+        .iter()
+        .filter(|span| !is_terminal(span.node_id))
+        .map(|span| span.node_id)
+        .collect()
+}
+
+pub(crate) fn build_unified_diff_graph(
```
dkhofer, line 721: It'd be great to have a high-level comment that walks through what this function does

gen-diff/src/graph.rs
dkhofer: Could you add a TODO comment to that effect?

gen-diff/src/graph.rs
```diff
@@ -338,145 +1153,4 @@ mod tests {
             .expect("graph edge");
         assert_eq!(weights, vec![edge.edge]);
     }
-
-    #[test]
-    fn get_diff_graph_splits_blocks_and_marks_new() {
```
dkhofer, line 343: Why is this getting removed?

gen-diff/src/operations.rs
dkhofer: I vaguely understand what's going on in the other file but I think there's enough there that it shouldn't be in model code.  It almost seems more appropriate for gen-graph if we want to put it somewhere else

gen-diff/src/operations.rs
```diff
@@ -1,796 +1,2066 @@
...
+/// Returns the affected block groups, their edge changes, and their operation attribution.
+///
+/// `build_dolt_operation_diff` consumes this aggregate to decide which block-group graphs to build
+/// and how to match and attribute edge changes within each graph.
+fn collect_block_group_changes(
+    graph_conn: &GraphConnection,
+    diff_ref: &str,
+) -> Result<BlockGroupChanges, OperationDiffError> {
+    let mut changes = BlockGroupChanges::default();
+    let block_group_query = format!(
```
dkhofer, line 295: I'm seeing multiple sql calls in this file, they should get pushed down to gen-models if at all possible

gen-diff/src/operations.rs
```diff
@@ -1,796 +1,2066 @@
...
+
+    struct DivergentHistory {
+        context: DbContext,
+        source_block_group: BlockGroup,
+        target_block_group: BlockGroup,
+        source_hash: DoltHashId,
+        target_hash: DoltHashId,
+    }
+
+    // Builds two branches with one unique block group on each side:
```
dkhofer, line 916: THIS is the explanation I was looking for for the two dot vs. three dot syntax, please put it near the top of this file

gen-diff/src/operations.rs
```diff
@@ -1,796 +1,2066 @@
...
+        Ok(edges)
+    }
+
+    fn load_augmented_edge(
+        graph_conn: &gen_models::db::GraphConnection,
+        block_group_id: HashId,
+        edge_key: BlockGroupEdgeKey,
+        history_ref: &str,
+    ) -> Result<AugmentedEdge, OperationDiffError> {
+        let query = "SELECT
```
dkhofer, line 1058: Another sql query

gen-models/migrations/config/01-initial/up.sql
dkhofer: :+1:

dkhofer: Maybe also rename the folder from `operations` to `config`?

gen-models/src/block_group.rs
```diff
@@ -773,32 +810,42 @@ impl BlockGroup {
...
     fn persist_insert_changes(
         conn: &GraphConnection,
         new_augmented_edges_by_block_group: HashMap<HashId, Vec<AugmentedEdgeData>>,
         new_accession_edges: HashMap<(HashId, String), Vec<AugmentedEdgeData>>,
     ) -> Result<(), BlockGroupError> {
-        let mut edge_data_map = HashMap::new();
-
         for (block_group_id, new_augmented_edges) in new_augmented_edges_by_block_group {
-            let new_edges = new_augmented_edges
+            let mut unique_new_edges = new_augmented_edges
```
dkhofer, line 823: Codex tells me the changes here are to ensure the things being inserted are de-duped, and are in the right order, and this helps doltlite to track object IDs for its history records.  Is that basically right?

gen-models/src/operations.rs
```diff
@@ -3024,111 +1398,125 @@ mod tests {
...
+                .iter()
+                .all(|asset| asset.role == AssetRole::Input)
+        );
+    }
+
+    #[test]
+    fn test_add_files_operation_records_history_without_config_db_rows() {
+        let context = setup_gen();
+        let graph_conn = context.graph().conn();
+        let operation_conn = context.config().conn();
```
dkhofer: This should be named config_conn

gen-r/R/genplot_widget.R
```diff
@@ -6,6 +6,13 @@
 }
 
 .genplot_widget <- function(frame_json) {
+  if (!requireNamespace("anyhtmlwidget", quietly = TRUE)) {
```
dkhofer, line 10: This is kind of random, what's the reason for this change?

gen-r/vignettes/introduction.Rmd
```diff
@@ -14,6 +14,16 @@ editor_options:
 
 ```{r, include = FALSE}
 knitr::opts_chunk$set(eval = TRUE)
+has_anyhtmlwidget <- requireNamespace("anyhtmlwidget", quietly = TRUE)
```
dkhofer, line 17: Same question as above, why the need for this change?

gen-tui/src/edge_router/route_graph.rs
```diff
@@ -142,10 +142,9 @@ pub fn make_rectilinear(
                     let bundle = if let Some(edge_idx) = graph.find_edge(*node_index1, *node_index2)
                     {
                         graph.edge_weight(edge_idx).unwrap().bundle.clone()
-                    } else if let Some(edge_idx) = graph.find_edge(*node_index2, *node_index1) {
-                        graph.edge_weight(edge_idx).unwrap().bundle.clone()
                     } else {
-                        return None; // No edge exists
```
dkhofer, line 148: Why is this getting deleted?

src/commands/import/library.rs
```diff
@@ -79,19 +82,19 @@ pub fn execute(cli_context: &CliContext, cmd: Command) -> Result<()> {
         Some(&parts_path),
         Some(&library_path),
     ) {
-        Ok(_) => {
-            println!("Imported library file {library_path} and parts file {parts_path}");
-            conn.execute("END TRANSACTION;", []).unwrap();
```
dkhofer, line 84: Why is it necessary to begin and rollback transactions, but no longer necessary to end them?

src/commands/remote/client.rs
```diff
@@ -0,0 +1,876 @@
...
+        let repository_segments = match segments.as_slice() {
+            ["repos", namespace, slug] | ["api", "repos", namespace, slug] => (*namespace, *slug),
+            _ => {
+                return Err(RemoteClientError::InvalidRepositoryUrl(
+                    remote_url.to_string(),
+                ));
+            }
+        };
+        let namespace = repository_segments.0.to_string();
+        let slug = repository_segments.1.trim_end_matches(".git").to_string();
```
dkhofer: Do we really need to worry about .git suffixes?

src/views/diff.rs
dkhofer: :+1:

src/views/diff_graph.rs
dkhofer: :+1:

src/lib.rs
dkhofer: :+1:

src/main.rs
```diff
@@ -81,36 +72,111 @@ fn call_cli() -> Result<(), Box<dyn std::error::Error>> {
...
+                "{head_marker:<3}{hash:>40}   {summary:<70}",
+                hash = entry.commit_hash,
+                summary = entry.message
+            );
+        }
+        return Ok(());
+    }
 
-    let operation_conn = get_operation_connection(Some(workspace.gen_db_path()?))?;
+    let operation_conn = get_config_connection(Some(workspace.gen_db_path()?))?;
```
dkhofer: Should probably do a global s/operation_conn/config_conn/

dkhofer (approved):
There's a lot here, but it all seems reasonable.  I had a bunch of minor comments but none are blocking.  LGTM!

gen-tui/src/edge_router/route_graph.rs
Chris7: new code is equivalent, it adds a ? so the None branch returns automatically

src/commands/import/library.rs
Chris7: commit_operation below ends it, but i agree it's confusing having it happen outside the thing that initiated it.

gen-models/src/block_group.rs
Chris7: essentially, the sort is also here to improve insertions into a btree

PR #220, opened 2026-07-29
Pip cli

bobvh:
The idea was to make installing gen as easy and non-threatening as possible.
Most users will be comfortable with a pip install, and you can include an executable binary alongside the python bindings.

I changed as little as possible, but did add a few extra builds to get to a crossplatform solution and not have it fail depending on your OS or python install. The abi3 switch was because the number of python builds was getting out of hand, I could revert it if needed.

tl;dr:
 - added a windows-release job that also uploads an asset.
 - added intel for mac and arm for linux builds
 - macOS binaries are codesigned
 - add gen-python/scripts/verify_wheel.py to run in CI
 - switched python build to abi3 (one wheel for all python versions)
 - updated the install section of the readme

### DIFF
--- a/.github/workflows/python-wheels.yaml
+++ b/.github/workflows/python-wheels.yaml
@@ -39,9 +39,7 @@ jobs:
         if: runner.os == 'macOS'
         run: make stage-python-client
       - name: Sign bundled client (macOS)
-        # TEMPORARY: dropped the `github.ref_type == 'tag'` gate to verify signing
-        # works in CI without cutting a release tag. Restore the gate before merging.
-        if: runner.os == 'macOS'
+        if: runner.os == 'macOS' && github.ref_type == 'tag'
         env:
           MACOS_CERTIFICATE: ${{ secrets.PROD_MACOS_APP_CERTIFICATE }}
           MACOS_CERTIFICATE_PWD: ${{ secrets.PROD_MACOS_APP_CERTIFICATE_PWD }}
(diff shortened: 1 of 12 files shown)


.github/workflows/python-wheels.yaml
```diff
@@ -11,31 +11,101 @@ jobs:
...
-          python-version: ${{ matrix.python-version }}
+          python-version: "3.11"
       - uses: actions-rust-lang/setup-rust-toolchain@v1
       - name: Install capnp (macOS)
         if: runner.os == 'macOS'
         run: brew install capnp
       - name: Install capnp (Windows)
         if: runner.os == 'Windows'
         run: choco install capnproto
+      - name: Stage bundled client (macOS)
```
bobvh, line 38: This is something I wasn't able to test properly @dkhofer , how did you test this when you were doing this for the pkg?

bobvh: Learned more and promptly broke it, please disregard.

bobvh: Going to temporarily run it from the PR this one time to confirm that it works. It shouldn't push anything to pypi.

bobvh: Confirmed: https://github.com/genhub-bio/gen/actions/runs/30555111353/job/90913458025

Cargo.toml
```diff
@@ -25,6 +25,11 @@ members = [".", "gen-core", "gen-models", "gen-graph", "gen-diff", "gen-tui", "g
 default-members = [".", "gen-core", "gen-models", "gen-graph", "gen-diff", "gen-capnp-schemas", "gen-annotations"]
 exclude = ["gen-python", "gen-r/src/rust"]
 
+[patch.crates-io]
+# TEMPORARY: verifying the Windows MSVC build fix in CI before it's published
+# to crates.io. Revert this patch before merging.
```
bobvh, line 30: Only revert this if PR2 on rusqdoltlite has merged and published.

bobvh:
Superseded — broke this up into smaller units for review, continued as pip-client (#226) → pip-abi3 (#227) → pip-targets (#228) → pip-apple-signing (#230).

PR #221, opened 2026-07-29
Fix annotations hiding across variant bubbles in truncated view

bobvh:
During the demo we noticed fewer annotations being shown than expected.

The TUI in truncated mode is supposed to only show annotations that span an edge, or cover every position of graphnode. That's `what span_is_single_node` and `span_should_hide_in_trunated `are about.

The bug was that node in this context refers to the GraphNode in a GenGraph, not Node from the db.

This fixes that and clarifies the text on screen so the user know better what is shown and what is hidden

Split #222 fixes unrelated auto-activation fix found during the same demo.

### DIFF
--- a/gen-python/src/python_api/jupyter_widget.rs
+++ b/gen-python/src/python_api/jupyter_widget.rs
@@ -525,7 +525,7 @@ impl GraphPage {
             let note = if detail_level == VisualDetail::Full {
                 " some annotations hidden due to space constraints "
             } else {
-                " some annotations hidden in truncated view "
+                " only annotations spanning a variant (edge) or filling a full node are shown "
             };
             let theme = current_theme();
             let note_style = Style::default().fg(theme[0x09]).bg(theme[0x00]);
--- a/src/views/annotation_track.rs
+++ b/src/views/annotation_track.rs
@@ -54,13 +54,6 @@ pub fn annotation_span_from_graph_locus(locus: &GraphLocus, name: &str) -> Annot
 }
 
 /// Map `span`'s per-node segments onto the current graph.
-///
-/// A `node_id` identifies a parent `Node`, not a single slice of it: an edit that splits
-/// a node (a library insertion, a sequence update, ...) leaves several `GraphNode`s in the
-/// current graph sharing that same `node_id` but covering different, disjoint sequence
-/// ranges. Each segment is matched to whichever of those candidate slices its own range
-/// overlaps, rather than to an arbitrary one, so a same-`node_id` segment always lands on
-/// the fragment it actually belongs to.
 pub fn graph_locus_from_annotation_span(
     span: &AnnotationSpan,
     graph: &GenGraph,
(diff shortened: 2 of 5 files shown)


gen-python/src/python_api/jupyter_widget.rs
```diff
@@ -525,7 +525,7 @@ impl GraphPage {
             let note = if detail_level == VisualDetail::Full {
                 " some annotations hidden due to space constraints "
             } else {
-                " some annotations hidden in truncated view "
```
bobvh, line 532: This was too vague, especially since none were showing

src/views/annotation_track.rs
```diff
@@ -54,13 +54,6 @@ pub fn annotation_span_from_graph_locus(locus: &GraphLocus, name: &str) -> Annot
 }
 
 /// Map `span`'s per-node segments onto the current graph.
-///
-/// A `node_id` identifies a parent `Node`, not a single slice of it: an edit that splits
-/// a node (a library insertion, a sequence update, ...) leaves several `GraphNode`s in the
-/// current graph sharing that same `node_id` but covering different, disjoint sequence
-/// ranges. Each segment is matched to whichever of those candidate slices its own range
-/// overlaps, rather than to an arbitrary one, so a same-`node_id` segment always lands on
-/// the fragment it actually belongs to.
```
bobvh, line 63: Had left this in in attempt at making the agents not make the same mistake about node_id not being a unique identifier, but in the bug we're fixing here it clearly did do that again.

src/views/block_group.rs
```diff
@@ -198,6 +199,86 @@ fn load_annotation_groups_for_viewport(
     }
 }
 
+/// Shared inputs for loading an annotation file's track, common to activating one file
```
bobvh: I'll split this out into a separate PR

gen-python/src/python_api/jupyter_widget.rs
dkhofer: Yes, 👍

src/views/annotation_track.rs
dkhofer: 👍

src/views/annotation_track.rs
```diff
@@ -126,24 +122,20 @@ pub fn span_is_single_node(span: &AnnotationSpan) -> bool {
...
-        .find(|node| node.node_id == segment.node_id)
-    else {
+    let Some(first) = locus.slices.first() else {
         return false;
     };
-    !(segment.start <= node.sequence_start && segment.end >= node.sequence_end)
+    if !locus.slices.iter().all(|slice| slice.block == first.block) {
+        return false;
+    }
+    !(first.start == 0 && first.end as i64 >= first.block.length())
```
dkhofer, line 138: I think maybe all this could be simplified to `!span_is_single_node(span, graph) || !(first.start == 0 && first.end as i64 >= first.block.length())` ?  I'm also a little confused because there's sort of a double negative in the logic, the method is `should_hide` and then returns false for the cases it cares about, might be simpler to call the method `should_show_in truncated` and then return true?

dkhofer (approved):
LGTM.  Just a non-blocking suggestion for logic cleanup, feel free to disregard if I'm mistaken about the intent

src/views/annotation_track.rs
bobvh: Done

PR #223, opened 2026-07-29
Patch operation asset fix

Chris7:
The overall change is making the checksum optional in OperationFileInfo. It's currently required in the software but not by the data model. A few things were fixed/combined to make it so checksums could be calculated at the same time as data was being processed. That is the reason for the deletion of checksum/copy_to_local_path/ensure_file/etc. functions. It's centralized in stage_file_asset so we can copy a file and get the checksum at the same time.

The other consideration was dealing with remote uris. Currently we were downloading them just to checksum. I removed this so we don't do downloads just for this. It's the caller's job to download and checksum if that's the only goal.

### DIFF
--- a/gen-models/src/assets.rs
+++ b/gen-models/src/assets.rs
@@ -99,10 +99,6 @@ fn opendal_file_addition_error(err: opendal::Error) -> FileAdditionError {
     FileAdditionError::FileReadError(io::Error::other(err))
 }
 
-fn opendal_file_store_error(err: opendal::Error) -> FileStoreError {
-    FileStoreError::IoError(io::Error::other(err))
-}
-
 #[derive(Clone, Debug, Eq, Hash, PartialEq)]
 pub enum AssetRole {
     Input,
@@ -624,10 +620,6 @@ impl ChecksumHandle {
         let state = self.state.lock().unwrap();
         if state.complete { state.checksum } else { None }
     }
-
-    pub fn finalized_checksum(&self) -> Option<Sha256Hash> {
-        self.state.lock().unwrap().checksum
-    }
 }
 
 pub struct ChecksummedReader {
@@ -652,32 +644,28 @@ impl ChecksummedReader {
             state: Arc::clone(&self.state),
         }
     }
-}
 
-impl Read for ChecksummedReader {
-    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
-        let bytes_read = self.inner.read(buf)?;
-        let mut state = self.state.lock().unwrap();
-        if bytes_read == 0 {
-            if state.checksum.is_none() {
-                let finalized = state.hasher.clone().finalize();
-                state.checksum = Some(Sha256Hash(finalized.into()));
-            }
-            state.complete = true;
-        } else {
-            state.hasher.update(&buf[..bytes_read]);
-        }
-        Ok(bytes_read)
-    }
-}
-
-impl Drop for ChecksummedReader {
-    fn drop(&mut self) {
+    fn finish(&self) {
         let mut state = self.state.lock().unwrap();
         if state.checksum.is_none() {
             let finalized = state.hasher.clone().finalize();
             state.checksum = Some(Sha256Hash(finalized.into()));
         }
+        state.complete = true;
+    }
+}
+
+impl Read for ChecksummedReader {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        let bytes_read = self.inner.read(buf)?;
+        if bytes_read == 0 {
+            if !buf.is_empty() {
+                self.finish();
+            }
+        } else {
+            self.state.lock().unwrap().hasher.update(&buf[..bytes_read]);
+        }
+        Ok(bytes_read)
     }
 }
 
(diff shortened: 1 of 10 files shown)


gen-models/src/assets.rs
```diff
@@ -496,35 +496,33 @@ impl ChecksummedReader {
...
-            state.complete = true;
         } else {
-            state.hasher.update(&buf[..bytes_read]);
+            self.state.lock().unwrap().hasher.update(&buf[..bytes_read]);
         }
         Ok(bytes_read)
     }
 }
 
-impl Drop for ChecksummedReader {
```
Chris7, line 674: this isn't useful, storing the checksum in an object that is dropped doesn't matter since the object is gone.

gen-models/src/assets.rs
```diff
@@ -609,82 +607,30 @@ impl OpenDalLocation {
             .map_err(opendal_file_addition_error)?;
         Ok(ChecksummedReader::new(Box::new(reader)))
     }
-
-    fn checksum(&self, display_path: &str) -> Result<Sha256Hash, FileAdditionError> {
```
Chris7, line 769: we combined this in another place to prevent reading a remote file just for a checksum

gen-models/src/assets.rs
```diff
@@ -609,82 +607,30 @@ impl OpenDalLocation {
...
-                io::ErrorKind::NotFound => Ok(Sha256Hash::convert_str("non-existent")),
-                io::ErrorKind::PermissionDenied => Err(FileAdditionError::FilePermissionDenied(
-                    display_path.to_string(),
-                )),
-                _ => Err(FileAdditionError::FileReadError(err)),
-            },
-        }
-    }
-
-    fn copy_to_local_path(
```
Chris7, line 797: removed functions using this

gen-models/src/assets.rs
```diff
@@ -853,18 +797,6 @@ impl AssetUri for LocalAssetUri {
         Ok(Self::asset_uri(&relative_file_path))
     }
 
-    fn ensure_asset(
```
Chris7, line 1012: removed functions using this

gen-models/src/assets.rs
```diff
@@ -1147,6 +1079,54 @@ impl LocalAssetUri {
...
+        if let Some(checksum) = checksum_override {
+            let asset_path = asset_dir.join(self.asset_filename(&checksum));
+            if asset_path.exists() {
+                return Ok(checksum);
+            }
+        }
+
+        let mut reader = self.reader(workspace)?;
+        let checksum_handle = reader.checksum_handle();
+        let mut staged_file = tempfile::NamedTempFile::new_in(&asset_dir)
```
Chris7, line 1256: we have a staged file here so we don't end up with incomplete copies

gen-models/src/assets.rs
```diff
@@ -1192,54 +1172,35 @@ impl LocalAssetUri {
...
-            .map_err(FileAdditionError::FileReadError)?;
-        Ok(())
-    }
-
     pub fn store_file(
         file_addition: &FileAddition,
         workspace: &Workspace,
     ) -> Result<(), FileStoreError> {
+        let checksum = file_addition.checksum.ok_or_else(|| {
+            FileStoreError::IoError(io::Error::new(
```
Chris7, line 1336: i was wondering if this should be a soft fail. The goal is effectively to prevent local file reads being done w/o using the checksum'd variant and passing it along.

gen-models/src/assets.rs
```diff
@@ -1342,10 +1297,6 @@ impl RemoteAssetUri {
             asset_uri: asset_uri.to_string(),
         }
     }
-
-    fn checksum_for_uri(asset_uri: &str, checksum_override: Option<Sha256Hash>) -> Sha256Hash {
```
Chris7, line 1502: we don't have checksums for uris that are urls. The caller needs to decide to read the file and store it.

gen-models/src/operations.rs
```diff
@@ -59,13 +59,7 @@ pub struct OperationFileInfo {
     pub file_path: String,
     pub asset_uri: String,
     pub file_type: FileTypes,
-    pub checksum: Sha256Hash,
```
Chris7, line 62: this is essentially where these fixes came from. The serve rwas getting a 500 since some checksums were None.

gen-models/src/patch.rs
```diff
@@ -118,13 +118,7 @@ pub fn operation_asset_files_for_logs(
                 file_path,
                 asset_uri,
                 file_type,
-                checksum: row.get::<_, Option<Sha256Hash>>(3)?.ok_or_else(|| {
```
Chris7, line 121: this is where the other errors came from, this refuses to deserializes None types.

gen-models/src/assets.rs
dkhofer: :+1:

gen-models/src/assets.rs
dkhofer: :+1:

gen-models/src/assets.rs
dkhofer: :+1:

gen-models/src/assets.rs
dkhofer: :+1:

gen-models/src/assets.rs
dkhofer: :+1:

gen-models/src/assets.rs
dkhofer: I guess the alternative would be to let `stage_asset_copy` compute the checksum below?  I guess I'm fine with seeing if the logic here causes problems

gen-models/src/operations.rs
dkhofer: :+1:

gen-models/src/patch.rs
dkhofer: :+1:

gen-models/src/assets.rs
dkhofer: Actually, I'm not OK with this, if I'm reading things correctly.  The new stage_asset_copy method copies to the asset directory, but this method is the one that copies out to the repo location.  I'm not seeing anything in the new code that ensures the asset file is copied to the place the user would expect to see it.  Am I missing something?

In any case, once this gets to genhub, would be good to test it on staging

dkhofer (commented):
Mostly looks good, but I want to understand the changes to the code that does asset manifestation

gen-models/src/assets.rs
Chris7: Download_asset handles that as part of remote operations, but one thing I need to verify is whether checkout/reset/etc. restores files

Chris7: This also put stuff into .assets though. ensure_asset_copy below copies stuff to the .asset dir too. I don't think there's repo materialization happening here either.

gen-models/src/assets.rs
Chris7: yeah, it's a case where the answer is "use checksummedreader".

PR #224, opened 2026-07-29
Asset access update

Chris7:
This was some parallel improvements related to other branches. This changes the way we handle file readers so seek behavior can use OpenDal's native methods. This is done by changing from a ChecksummerReader -> blocking::StdReader. This was added because for remote annotation files, we want to be able to seek when using an index.

The s3 stuff was added related to that as well, where we will fail back to anonymous access instead of using credentials.

### DIFF
--- a/gen-models/src/operations.rs
+++ b/gen-models/src/operations.rs
@@ -802,6 +802,7 @@ mod tests {
         fs,
         io::{Cursor, Write},
         path::PathBuf,
+        slice::from_ref,
     };
 
     use tempfile::NamedTempFile;
@@ -1422,7 +1423,7 @@ mod tests {
 
         add_files_operation(
             &context,
-            std::slice::from_ref(&operation_file),
+            from_ref(&operation_file),
             Some("track remote reference"),
         )
         .expect("should track remote asset");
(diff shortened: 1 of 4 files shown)


docs/commands.md
```diff
@@ -36,6 +36,21 @@ Operations are changes that have been made to the database. Commands such as `im
 To see all operations, `gen operations` will list operations. The operation the database currently is on
 will be prefixed with a `>`.
 
+# Cache
```
Chris7: Cache added for temporary downloads. This is used for viewing/using remote urls in cases where we can't stream data from them and need a local copy

gen-models/src/operations.rs
```diff
@@ -802,6 +802,7 @@ mod tests {
         fs,
         io::{Cursor, Write},
         path::PathBuf,
+        slice::from_ref,
```
Chris7, line 805: i asked it to fix fully qualified imports and this drug along too

dkhofer: :+1:

dkhofer (approved):
LGTM

PR #226, opened 2026-07-30
Bundle the gen CLI client in the Python wheel

bobvh:
Stages the release Rust binary into maturin's wheel data `scripts` directory so `pip`, `uv`, and `pipx` installs put `gen` on PATH alongside the Python bindings.

- `make stage-python-client` builds the binary and copies it into `gen.gen.data/scripts/`
- Wheel builds on macOS, Linux, and Windows stage the client before packaging
- `gen-python/scripts/verify_wheel.py` checks the built wheel contains the client, the compiled extension, and the Jupyter widget asset
- CI installs the built wheel and runs `gen --version` as a smoke test
- README / gen-python README updated to describe the bundled client and new `make python-wheel` target

Also includes a temporary `[patch.crates-io]` pin of `rusqdoltlite` to an unpublished fix-windows branch (bobvh/rusqdoltlite@d1ef39b) — without it, `cargo build --bin gen` fails on Windows under MSVC (`pthread_t`/`DoltliteServer` parse failure in the bundled remote-auth server). Revert once the fix ships on crates.io.

Also bumps `spin` to 0.9.9 to resolve a yanked-crate warning from cargo-deny.

Stacked PR 1/5: pip-client → pip-abi3 → pip-targets → pip-publish-gating → pip-apple-signing


.github/workflows/python-wheels.yaml
```diff
@@ -27,13 +27,23 @@ jobs:
...
+          cargo build --release --locked --bin gen
+          New-Item -ItemType Directory -Force "gen.gen.data/scripts"
+          Copy-Item "target/release/gen.exe" "gen.gen.data/scripts/gen.exe"
       - name: Build wheels (Linux)
         if: runner.os == 'Linux'
         uses: PyO3/maturin-action@v1
         with:
           command: build
           args: --release --manifest-path gen-python/Cargo.toml --features extension-module --interpreter python${{ matrix.python-version }} --out gen-python/target/wheels
-          sccache: true
```
Chris7, line 36: was this failing?

.github/workflows/python-wheels.yaml
```diff
@@ -58,6 +69,31 @@ jobs:
           args: --release --manifest-path gen-python/Cargo.toml --features extension-module --interpreter ${{ steps.setup-python.outputs.python-path }} --out gen-python/target/wheels
           sccache: true
           working-directory: .
+      - name: Verify wheel contents (macOS, Linux)
+        if: runner.os != 'Windows'
+        run: python gen-python/scripts/verify_wheel.py gen-python/target/wheels/*.whl
```
Chris7, line 74: i don't really get the point of a verify_wheel script, what are we trying to keep track of?

gen-python/scripts/verify_wheel.py
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
+"""Verify that a Gen wheel contains the Python package and bundled client."""
```
Chris7, line 2: why do we need this?

Chris7: this doesn't really do a meaningful test. It's like validating a mock you wrote. Meaningful test is verifying it installs + is importable + is a callable binary.

Makefile
```diff
@@ -1,8 +1,17 @@
-.PHONY: python jupyter r r-test release-check-js clean build clippy-fix docker-build gif
+.PHONY: python python-wheel stage-python-client jupyter r r-test release-check-js clean build clippy-fix docker-build gif
 python:
 	@[ -d .venv ] || python -m venv .venv
 	@.venv/bin/pip show maturin >/dev/null 2>&1 || .venv/bin/pip install maturin
 	.venv/bin/maturin develop --release --manifest-path gen-python/Cargo.toml --features extension-module
+stage-python-client:
```
Chris7, line 6: why is it called stage-python-client?

README.md
```diff
@@ -8,19 +8,33 @@ Gen brings version control to genetic sequences. With it, you can track variants
 
 ## Install
 
-**Gen client**: prebuilt binaries for macOS and Linux are on the [releases page](https://github.com/genhub-bio/gen/releases): [macOS (.pkg)](https://github.com/genhub-bio/gen/releases/download/nightly/gen.macos.pkg), [Linux x86_64 (.zip)](https://github.com/genhub-bio/gen/releases/download/nightly/gen.linux-x86_64.zip), [Linux arm64 (.zip)](https://github.com/genhub-bio/gen/releases/download/nightly/gen.linux-arm64.zip). Gen is built primarily for Unix-like systems; on Windows, you can install [WSL](https://learn.microsoft.com/en-us/windows/wsl/) to get a Linux environment, then use the Linux binary above from inside it.
```
Chris7, line 11: i'd prefer to keep the binary stuff at the top -- imo that's the easiest way to install

.github/workflows/python-wheels.yaml
bobvh: The order of operations was wrong (sccache wasn't available at build time), disabling it was a bandaid, properly fixed now and re-enabling.

.github/workflows/python-wheels.yaml
bobvh: It wasn't absolutely necessary, I'll assume that if other tests pass the packaging will proceed normally. Removing the script altogether.

gen-python/scripts/verify_wheel.py
bobvh: Removing it completely.

Makefile
bobvh: It was staging the CLI in the pip package. Renamed to py-package-cli to make it more clear.

README.md
bobvh: Moving it back to the top.

PR #227, opened 2026-07-30
Build the Python extension against the CPython stable ABI

bobvh:
Switches to pyo3's `abi3-py311` feature so one wheel per OS covers every supported Python version, instead of building once per interpreter (which was about to balloon from 3 platforms × 4 versions to 5 platforms × 4 versions once targets are added).

- `gen-python` builds with `--features abi3,extension-module` against Python 3.11 only
- To make sure it is still compatible with 3.12-3.14 the `test-newer-python` job installs it with those versions
- `verify_wheel.py` now also checks the wheel is tagged `-abi3-`

Stacked PR 2/5: pip-client → pip-abi3 → pip-targets → pip-publish-gating → pip-apple-signing

### DIFF
--- a/.github/workflows/python-wheels.yaml
+++ b/.github/workflows/python-wheels.yaml
@@ -12,14 +12,13 @@ jobs:
       fail-fast: true
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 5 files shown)


Chris7:
lgtm

PR #228, opened 2026-07-30
Add Linux arm64, macOS Intel, and Windows wheel targets

bobvh:
Expands the wheel build matrix with `ubuntu-24.04-arm` and `macos-15-intel`, and adds a `release-windows` job so nightly/tagged GitHub releases ship a Windows zip alongside the existing macOS and Linux binaries (otherwise they'd only be in the pip install route).

- `python-wheels.yaml` matrix grows to 5 platforms
- `release.yml` gets a `release-windows` job building `gen.exe` and uploading `gen.windows-x86_64.zip`
- README install section links the Windows release asset

Stacked PR 3/5: pip-client → pip-abi3 → pip-targets → pip-publish-gating → pip-apple-signing

### DIFF
--- a/.github/workflows/python-wheels.yaml
+++ b/.github/workflows/python-wheels.yaml
@@ -11,7 +11,12 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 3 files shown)


Chris7:
lgtm

PR #229, opened 2026-07-30
Gate PyPI publish on a specific tested wheel-build run

bobvh:
Wheels are now built/tested every PR and push to `main`, uploading to the GitHub release and triggering PyPI publish, including code signing (see next PR) still only happen for tag-triggered runs.

The old publishing system did a lookup for wheels made for a given commit, but not all of those are signed (next PR) so that's why `trigger-pypi-publish` now passes its own run ID explicitly, and `publish-pypi-tag.yaml` validates that exact run before downloading artifacts from it:

- correct workflow (`python-wheels.yaml`)
- completed successfully
- its `upload-release-wheels` job succeeded
- the given tag actually points at that run's commit

Stacked PR 4/5: pip-client → pip-abi3 → pip-targets → pip-publish-gating → pip-apple-signing

### DIFF
--- a/.github/workflows/publish-pypi-tag.yaml
+++ b/.github/workflows/publish-pypi-tag.yaml
@@ -1,13 +1,15 @@
 name: Publish PyPI Tag
 
 on:
-  push:
-    tags:
-      - "v*"
+  # Python wheels dispatch this workflow only after release gating succeeds.
   workflow_dispatch:
     inputs:
       tag:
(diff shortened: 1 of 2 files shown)


Chris7 (commented):
lukewarm on this change, why do we need to change how we publish?

bobvh:
@Chris7 you're right, this was over optimizing on something that wasn't that big of an issue. I'm dropping this PR altogether from the chain.

PR #231, opened 2026-07-31
Asset commit range query

Chris7:
Before this change, every asset transfer included the branch’s entire file history. This gives a way to scope it to refs. It also gives 2 ways to do the call -- get every version of the file in that range (called cumulative here), or just the last version (called materialized here).

The transfer code was updated to use these new methods along with some expanded test cases

### DIFF
--- a/gen-models/src/assets.rs
+++ b/gen-models/src/assets.rs
@@ -316,7 +316,49 @@ impl AssetRef {
         to_hash: Option<&DoltHashId>,
         view: AssetView,
     ) -> Result<Vec<Self>, QueryError> {
-        let query = "WITH RECURSIVE ancestry(commit_hash, depth) AS ( \
+        let (ancestry_boundary, history_depth, view_query) = match view {
+            AssetView::Cumulative => (
+                "AND (:from_hash IS NULL OR ancestry.commit_hash <> :from_hash)",
+                "",
+                "SELECT id, uri, file_type, checksum, size, role, logical_path, name, created_on \
+                 FROM asset_versions \
+                 ORDER BY logical_path, id",
+            ),
+            AssetView::Materialized => (
+                "",
+                ", MAX(bounded_ancestry.depth) AS introduction_depth",
+                ", /* Keep only assets present at the upper bound, then rank each logical path by \
+                      its introduction order in first-parent history. Null paths remain \
+                      independent because they are partitioned by asset ID. */ \
+                 materialized_assets AS ( \
+                     SELECT asset_versions.*, \
+                            ROW_NUMBER() OVER ( \
+                                PARTITION BY asset_versions.logical_path, \
+                                             CASE \
+                                                 WHEN asset_versions.logical_path IS NULL \
+                                                 THEN asset_versions.id \
+                                                 END \
+                                ORDER BY asset_versions.introduction_depth, \
+                                         asset_versions.id \
+                            ) AS materialized_rank \
+                     FROM asset_versions \
+                     JOIN dolt_at_gen_asset_refs( \
+                         COALESCE(:to_hash, dolt_hashof('HEAD')) \
+                     ) AS current_assets \
+                       ON current_assets.id = asset_versions.id \
+                 ) \
+                 SELECT id, uri, file_type, checksum, size, role, logical_path, name, created_on \
+                 FROM materialized_assets \
+                 WHERE materialized_rank = 1 \
+                 ORDER BY logical_path, id",
+            ),
+        };
+        let query = format!(
+            "WITH RECURSIVE \
+                     /* Walk first-parent history from the upper bound. Cumulative queries stop at \
+                        the inclusive lower bound; materialized queries continue to the root so \
+                        version precedence comes from commit order rather than wall-clock time. */ \
+                     ancestry(commit_hash, depth) AS ( \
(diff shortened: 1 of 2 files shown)


gen-models/src/assets.rs
```diff
@@ -161,6 +162,11 @@ impl FromSql for AssetRole {
     }
 }
 
+/// An immutable reference to an asset recorded in repository history.
+///
+/// Several references can have the same logical path because Gen retains every committed version.
+/// Use [`Self::get_cumulative_assets_at`] when that provenance is needed and
```
dkhofer, line 176: Could you add a bit more to this comment here saying cumulative and materialized mean?  To me, cumulative implies everything from the first commit on, but maybe it also means all versions of a given file if it has gone through changes

gen-models/src/assets.rs
```diff
@@ -1476,34 +1595,277 @@ mod tests {
...
 
-        let assets = Assets::get_branch_assets(conn, &commit.to_string())
-            .expect("should read assets at commit");
+        let assets =
+            Assets::get_branch_assets(conn, "main").expect("should read assets from branch head");
 
-        assert_eq!(assets, HashMap::from([(local_asset.id, local_asset)]));
+        assert_eq!(
+            assets,
+            HashMap::from([(replacement_asset.id, replacement_asset)])
```
dkhofer, line 1920: I'm confused why this only includes the replacement asset.  I'm not seeing an indication that a new branch was made with only it having been added.  I'm also not sure why this asset should be considered a full replacement for the original asset, which appears to be different in all fields.  Is this behavior what's really expected?

src/commands/remote/operations.rs
```diff
@@ -361,15 +366,18 @@ fn download_asset(
...
-    let intended_change =
-        previous_asset
-            .and_then(|asset| asset.checksum)
-            .is_some_and(|previous_checksum| {
-                previous_checksum != expected_checksum
-                    && existing_checksum == Some(previous_checksum)
-            });
+    let intended_change = existing_checksum
+        .as_ref()
+        .map(|checksum| {
```
dkhofer: This seems to be more of the map/transpose slop we've seen before, is it possible to rework as "if destination_matches_previous_asset()...else..."?

src/commands/remote/operations.rs
```diff
@@ -278,39 +280,39 @@ enum DownloadAssetOutcome {
     Conflict(PathBuf),
 }
 
-/// This is effectively a dirty file check. On clones/pulls we want to update files if they
-/// match expected checksums. This gives the asset that was previously in a path so we can
-/// compare it
-fn previous_asset_for_destination<'asset>(
+/// This is effectively a dirty file check. On clones/pulls we want to update
+/// files if they match expected checksums. The previous asset set is cumulative
```
dkhofer: Shouldn't this say "we want to update files if they *don't* match expected checksums"?  The logic in the actual method looks fine, but this comment could be clearer

src/commands/remote/operations.rs
```diff
@@ -336,20 +338,23 @@ fn conflict_destination_path(
 /// On downloading an asset, if it is a file we compare the hash on disk to the hashes of the
 /// asset from previous updates. If it matches, we replace it assuming it's a natural evolution
```
dkhofer: I *think* this could be more accurately reworded as "If it matches the remote hash, we replace it locally assuming it's a natural evolution..."  Again, I think the code is fine, but this comment makes it hard to understand what it's doing

dkhofer (commented):
I'm kind of confused by some of the logic in this change (admittedly, the stuff causing my confusion may have been in there already).  It may be that I'm just misunderstanding some of the intended logic, but regardless I'd like to take another look assuming my comments are worth addressing

gen-models/src/assets.rs
Chris7: these have the same logical_path. That's where files end up in the filesystem. It's testing that if we have reference.fa and later update it and stil call it reference.fa, the replacement assets will be there.

src/commands/remote/operations.rs
Chris7: We want to update them if they match a known checksum on a pull, otherwise it would be overwriting potentially uncomitted work for users. This is what i mean by a dirty file check. It's like git pull and it saying you need to commit/stash changes first.

src/commands/remote/operations.rs
Chris7: sure, i can reword

gen-models/src/assets.rs
Chris7: well fortunately your intuition was right -- cumulative is all versions

dkhofer: Thanks

gen-models/src/assets.rs
dkhofer: Ah, got it.  I was reading "first.fa" in the uri and not seeing "reference.fa" as the logical path

src/commands/remote/operations.rs
dkhofer: Got it.  Could you just change the wording of this comment from "if they match expected checksums" to "if they match previous known checksums"?  That would have helped me

dkhofer (approved):
Thanks for your responses!  LGTM now

PR #232, opened 2026-07-31
Fix deletion bypass graph construction

### LINKED ISSUE #211: Gen update with sequence bug

Chris7:
Fixes #211

The stored data looked correct, so I treated it as a graph building problem. Effectively we these changes do 2 things:
* Make zero-width node so we can have connections to edges that affect the very beginning or end of sequence
* Use the same zero-width nodes for internal edges that are zero width as well. Zero width is detected by start_coordinate == end_coordinate for the edge. This is so we can represent processes that iteratively chunk off pieces of the front/end of a block. I also think this will show neighboring deletions better as individual deletions + a combined one.

Overall, this makes our layout and sequence stuff work just fine as it treats it as a junction point but renders it invisible. This seemed to render better than the previous attempt that only used edges and didn't introduce any blocks.

I don't love the variable names, if you have better/more obvious suggestions let me know.

image
image

### DIFF
--- a/gen-models/src/edge.rs
+++ b/gen-models/src/edge.rs
@@ -411,9 +411,17 @@ impl Edge {
         Ok(edges)
     }
 
+    /// Converts stored edge coordinates for one backing node into the sequence slices represented
+    /// by `GroupBlock`s.
+    ///
+    /// `blocks_from_edges` calls this after it has collected every coordinate needed by the block
+    /// group. Non-zero intervals carry sequence, while requested zero-width intervals act as
+    /// junctions between adjacent coordinate jumps. Outer junctions cover edge endpoints that have
+    /// no sequence interval on one side, so every stored edge still has a concrete graph endpoint.
     fn get_block_intervals(
         starts: &HashSet<i64>,
         ends: &HashSet<i64>,
+        junction_coordinates: &HashSet<i64>,
     ) -> Result<Vec<(i64, i64)>, EdgeError> {
         let coordinates = starts.union(ends).sorted().copied().collect::<Vec<_>>();
         if coordinates.is_empty() {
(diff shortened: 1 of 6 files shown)


gen-tui/src/graph_controller.rs
```diff
@@ -778,7 +778,9 @@ where
             .node_indices()
```
Chris7, line 778: everything from here on out is visual stuff making the zero-width nodes not selectable in the graph viewer.

Chris7:
I checked how main rendered side-by-side deletions and yeah, this seems better: 
image

is the unpatched version whereas the update shows a clear set of deletions side-by-side

gen-models/src/edge.rs
```diff
@@ -426,9 +434,83 @@ impl Edge {
...
+                .or_default()
+                .insert(edge.source_coordinate);
+            incoming_coordinates_by_node_id
+                .entry(edge.target_node_id)
+                .or_default()
+                .insert(edge.target_coordinate);
+        }
+    }
+
+    /// Materializes the backing-node slices needed to project stored block-group edges.
```
dkhofer: Suggestions for wording improvements: "Computes the backing-node slices from the stored block group edges."

gen-models/src/edge.rs
```diff
@@ -426,9 +434,83 @@ impl Edge {
...
+    ///
+    /// Graph construction, sequence enumeration, GFA export, and diff reconstruction use this as
```
dkhofer, line 483: More suggestions:

"Graph construction, sequence enumeration, GFA export, and diff reconstruction use this as the first half of the stored-edge-to'GenGraph' pipeline, the second half being the build_graph method below.  This method gathers node coordinates, expands incomplete node descriptions from the block group, and calls get_block_intervals to construct both blocks with sequences and zero-width junction blocks."

gen-models/src/edge.rs
```diff
@@ -535,7 +637,25 @@ impl Edge {
             let empty_ends = HashSet::new();
             let starts = starts_by_node_id.get(node_id).unwrap_or(&empty_starts);
             let ends = ends_by_node_id.get(node_id).unwrap_or(&empty_ends);
-            let block_intervals = Edge::get_block_intervals(starts, ends)?;
+            // Adjacent same-node jumps share a coordinate: one jump arrives at k and the next
+            // leaves from k. Materialize k as a junction so the graph connects both jumps:
```
dkhofer: Could just say "add" instead of "materialize" here

gen-models/src/edge.rs
```diff
@@ -566,15 +686,133 @@ impl Edge {
...
+
+        let source_blocks = Self::select_junction_or_sequence_blocks(source_blocks);
+        let target_blocks = Self::select_junction_or_sequence_blocks(target_blocks);
+        source_blocks
+            .into_iter()
+            .cartesian_product(target_blocks)
+            .collect()
+    }
+
+    /// Projects augmented stored edges and their materialized blocks into a `GenGraph`.
```
dkhofer: "Computes a GenGraph from augmented edges and their computed blocks"

gen-models/src/edge.rs
```diff
@@ -704,19 +991,190 @@ mod tests {
...
+        );
+        assert!(
+            !graph.contains_edge(graph_node(&blocks[0]), graph_node(&blocks[2])),
+            "outgoing edge should not bypass the junction"
+        );
+        assert_eq!(graph.edge_count(), 1, "should project one block edge");
+    }
+
+    #[test]
+    fn test_build_graph_projects_same_coordinate_edge_from_start_junction() {
```
dkhofer: I'm seeing the verb "project" or "projects" used a lot here, I think "create" or "creates" would work fine and be more understandable

gen-tui/src/graph_controller.rs
dkhofer: :+1:

dkhofer:
Could you also add a test to GFA export to see what happens there with a front-of-node deletion?

dkhofer:
Overall LGTM.  I'm good with "junction" or "junction block" as the term to describe the convenience/filler blocks

dkhofer (approved):

bobvh:
Sorry for the PR necromancy here, but there's a difference in meaning between the 2 versions of rendering here. It's the other PR that made me revisit this.

In the legacy image the visual indicates 3 possible routes: ATGA..., AGA..., ATA... but not AA... (no double deletions). That's what the dogleg bend shows since the rule is that you can only read from left to right. If the GenGraph actually had the edges that allow that route, the render would look like a small grid (two squares joined together on one side).

I'm not saying the new way is wrong per se, it's more of an information loss issue in that two distinct GFAs will have the same render, while before there was a 1:1 mapping.

PR #233, opened 2026-07-31
Fix annotation lookup in viewer

Chris7:
Fix the issue where child samples can't select annotations groups

### DIFF
--- a/src/views/annotations.rs
+++ b/src/views/annotations.rs
@@ -150,10 +150,12 @@ fn load_group_annotations(
     node_ids: &HashSet<HashId>,
     history_ref: Option<&str>,
 ) -> Result<Vec<AnnotationSpan>, AnnotationError> {
+    // The entry identifies the block group that owns the annotations. The current
+    // selection below is only the graph onto which those annotations are projected.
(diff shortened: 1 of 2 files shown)


src/views/annotations.rs
```diff
@@ -150,10 +150,12 @@ fn load_group_annotations(
     node_ids: &HashSet<HashId>,
     history_ref: Option<&str>,
 ) -> Result<Vec<AnnotationSpan>, AnnotationError> {
+    // Parent and ancestor entries retain the block group that owns their
+    // accessions; the current child graph below is still used for clipping.
     let annotations = Annotation::query_by_group_and_block_group(
         conn,
         &entry.name,
-        &current_block_group.id,
+        &entry.source_block_group_id,
```
Chris7, line 158: so this is the issue -- we were always looking up the current block group instead of the entry (the annotation track in question)'s block group. Hence the ancestral annotations never loaded b/c the bg id was wrong

dkhofer: Ah, nice find

dkhofer (approved):
LGTM, thanks for the fix!

PR #234, opened 2026-07-31
Update remote annotation access

Chris7:
So when using remote assets in the annotation viewer, the prior behavior was to download the entire thing. What should happen is:
* check to see if we have an index, if so, download it
* check to see if the server supports ranged reads
* used ranged reads if possible
* if no ranged reads are possible, download the annotation file to the cache and use it unless it already exists

So this reflects this behavior + adds a cache and cache-clear command to manage it

### DIFF
--- a/src/views/annotations.rs
+++ b/src/views/annotations.rs
@@ -1,11 +1,12 @@
 use std::{
     collections::{HashMap, HashSet},
     error::Error,
-    fs::File,
-    io::{BufRead, BufReader, Cursor},
+    fs::{self, File},
+    io::{self, BufRead, BufReader, Cursor, Read, Seek, Write},
     path::{Path as FsPath, PathBuf},
 };
 
+use flate2::read::MultiGzDecoder;
 use gen_annotations::{
     projection as annotation_projection,
     translate::{bed::translate_bed, gff::translate_gff},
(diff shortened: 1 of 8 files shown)


src/views/annotations.rs
```diff
@@ -391,26 +403,124 @@ pub fn parse_translated_bed<R: BufRead>(
...
-        .join(file_addition.hashed_filename()?);
+    let hashed_filename = file_addition.hashed_filename()?;
+    let asset_path = workspace.asset_dir().ok()?.join(hashed_filename);
     if asset_path.exists() {
         return Some(asset_path);
     }
     None
 }
 
+/// Builds the stable filename used when a remote annotation or index is materialized locally.
```
dkhofer: Another wording nitpick: Could just say "cache/cached" instead of "materialize/materialized" here and below

dkhofer (approved):
LGTM, very minor comment only.

I realize this is all being done to make the annotation viewing experience seamless, but out of curiosity, if the user knows they're going to be on a plane or something, is there a way for them to force download the annotation files so they have local copies of everything?

src/views/annotations.rs
Chris7: :+1:

Chris7:
There isn't a force download, it's a good idea though instead of making them go click through stuff over potentially hours

PR #235, opened 2026-08-02
Collection selector

Chris7:
Adds selector for collection in view command

### DIFF
--- a/gen-models/src/collection.rs
+++ b/gen-models/src/collection.rs
@@ -49,6 +49,19 @@ impl Query for Collection {
 }
 
 impl Collection {
+    /// Returns collections visible at the requested history state, ordered by name.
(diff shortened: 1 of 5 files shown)


gen-models/src/collection.rs
```diff
@@ -49,6 +49,19 @@ impl Query for Collection {
 }
 
 impl Collection {
+    /// Returns collections visible at the requested history state, ordered by name.
+    pub fn all(conn: &GraphConnection, history_ref: Option<&str>) -> Vec<Collection> {
```
Chris7, line 53: this needs to be updated at the trait level later

dkhofer: :+1:

dkhofer (approved):
LGTM

PR #237, opened 2026-08-03
Inline annotations deletion view

Chris7:
Fix the edges between inline deletions to retain the highlight:

image


src/views/gen_graph_widget.rs
```diff
@@ -919,6 +980,98 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_zero_width_junction_preserves_edge_highlight_glyph_and_color() {
```
bobvh, line 984: Yeah I wish the test backend / insta snapshots supported colors

bobvh (approved):
Looks good to me. This is only about the annotations, the complicated part was a different PR.

PR #239, opened 2026-08-04
Use complete_asset interface

Chris7:
This is added to tell the server what was uploaded and their CRC/md5sums. It's meant so a server can quickly verify that what was uploaded is what is actually in its storage. Its meant to prevent the server from having to do this work. This does not overwrite or set a storage checksum -- GCS/etc. calculates that itself. This verifies that what we thought was uploaded is what is actually uploaded. It's the final stage of a push currently.


src/commands/remote/operations.rs
```diff
@@ -257,18 +286,25 @@ fn upload_asset(
     let response = client
         .put(url)
         .header("content-type", "application/octet-stream")
+        .header("content-md5", &md5)
```
Chris7, line 293: for non-composite files (<5GB), this will be used as a server side validation the file was correctly received

src/commands/remote/client.rs
```diff
@@ -13,7 +13,9 @@
 //! runs the native Dolt operation, and then restores the canonical GenHub URL. If Dolt
 //! rejects an expired capability, the operation requests a fresh capability and retries.
 //! After the graph transfer, [`acquire_asset_transfers`] obtains the per-asset upload or
-//! download URLs used to transfer files referenced by the selected branch.
+//! download URLs used to transfer files referenced by the selected branch. A push then
+//! calls [`complete_asset_transfers`] with GCS-validated upload receipts so GenHub can
```
dkhofer, line 17: How tied to GCS is this implementation?

src/commands/remote/operations.rs
dkhofer: What happens for files > 5 GB?

dkhofer (approved):
Couple questions but LGTM

src/commands/remote/operations.rs
Chris7: gcs doesn't say. they might do what AWS does, but this isn't a check we use on the client/server side. It lets us do integrity validation if a file if < 5GB. Otherwise, google maybe tosses it out. We use the crc.

src/commands/remote/client.rs
Chris7: Not very, the receipt is just a crc32/md5sum the clinet sends. The sever can do what they wish with it.

PR #240, opened 2026-08-05
Refactor db graph code

dkhofer:
The gen-models crate currently wraps a couple pieces of functionality that this PR splits apart.  One piece is interactions with the database, which is what I think gen-models should be concerned with.  The other piece is taking the data loaded from the db, building a graph, and working with that graph.  I think that belongs elsewhere, for at least two reasons:

- The `get_all_sequences` method currently in the BlockGroup model is used in a bunch of tests in and outside of gen-models as a way to confirm graph structure.  That method relies on gen-graph to build the graph.  I think a crate primarily concerned with database interactions shouldn't have a dependency on a crate for graph building
- There are significant pieces of code within gen-models that don't have anything to do with the database.  For instance, graph.rs and region.rs each have hundreds of lines of code, but no sql statements (neither has a corresponding table in the database).

Currently the crate dependencies are:

gen-models -> gen-graph -> gen-core
                 \---------------------/

This PR changes the dependencies to:

gen-graph -> gen-models -> gen-core
                 \---------------------/

But I don't have a strong opinion about where the graph code should be extracted to, if we prefer a separate crate that's fine too.

Because gen-models tests still need to call `get_all_sequences`, which this PR moves to gen-graph, I set up a subcrate within gen-models for tests that need to use `get_all_sequences`, and moved them there.  That subcrate depends on both gen-graph and gen-models.  It also groups the tests together somewhat functionally.  I think that's an improvement because currently there are thousands of lines of test code in block_group.rs, which is somewhat unwieldy.

Since we keep extending the graph functionality code in gen-models, I thought it would be good to split it out now before it grows even more.


src/updates/vcf.rs
```diff
@@ -1147,7 +1172,7 @@ mod tests {
         let nodes = Node::query(conn, "select * from nodes;", rusqlite::params!());
         assert_eq!(nodes.len(), 5);
 
-        let second_update = update_with_vcf(
+        let _second_update = update_with_vcf(
```
dkhofer, line 1175: This is clippy

gen-core/src/graph.rs
```diff
@@ -0,0 +1,101 @@
+use core::fmt;
```
dkhofer, line 1: These got added here as part of the refactoring, to avoid circular dependencies.  I think they could get moved back to gen-models now but I don't have a strong opinion

gen-graph/src/graph_loader.rs
```diff
@@ -0,0 +1,659 @@
+use std::collections::{HashMap, HashSet, VecDeque};
```
dkhofer, line 1: I am open to better names for these files/modules and how the functions are organized within them, just wanted to extract this stuff out of models

gen-core/src/graph.rs
dkhofer: Er, guess they came from gen-graph originally, maybe better to have them here than gen-models

gen-core/src/graph.rs
```diff
@@ -0,0 +1,101 @@
...
+    pub edge_id: HashId,
+    pub source_strand: Strand,
+    pub target_strand: Strand,
+    pub chromosome_index: i64,
+    pub phased: i64,
+    pub created_on: i64,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct GraphLoadBlock {
```
Chris7, line 21: what are these for? Just doing a find in the GH diff view I just see them used for tests

gen-graph/src/graph_loader.rs
```diff
@@ -0,0 +1,659 @@
...
+                for node in path {
+                    sequence.push_str(&sequences_by_node[&node]);
+                }
+                sequences.insert(sequence);
+            }
+        }
+    }
+
+    sequences
+}
```
Chris7, line 659: any tests?

gen-graph/src/models.rs
```diff
@@ -0,0 +1,435 @@
...
+    let conn = context.graph().conn();
+    let parsed_region = gen_core::region::Region::parse(region)?;
+    let resolved_region = gen_models::region::resolve(&parsed_region, conn, collection, sample)?;
+    let tree = region_intervaltree(&resolved_region, conn)?;
+    let range = resolved_region.start..resolved_region.end;
+    let spans = AccessionSpan::from_intervaltree_ranges(&tree, core::slice::from_ref(&range))?;
+    persist_annotation(context, name, group, sample, &resolved_region, spans)
+}
+
+pub fn prune_graph(graph: &mut GenGraph) {
```
Chris7, line 43: is there a need for these indirect calls?

gen-graph/src/models.rs
```diff
@@ -0,0 +1,435 @@
...
+}
+
+pub fn get_all_sequences(
+    conn: &GraphConnection,
+    block_group_id: &HashId,
+) -> Result<HashSet<String>, BlockGroupError> {
+    get_all_sequences_with_pruning(conn, block_group_id, true)
+}
+
+pub fn get_all_sequences_with_pruning(
```
Chris7, line 106: was something failing and we needed this additional wrapper?

gen-graph/src/models.rs
```diff
@@ -0,0 +1,435 @@
...
+                    sequence_end: block.end,
+                },
+                block.sequence(),
+            )
+        })
+        .collect::<HashMap<_, _>>();
+    Ok(graph_loader::get_all_sequences(&graph, &sequences_by_node))
+}
+
+pub fn get_sample_all_sequences(
```
Chris7, line 137: get_all_sample_sequences sounds better to me

gen-graph/src/graph_loader.rs
```diff
@@ -0,0 +1,659 @@
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct RegionPositionQuery {
+    pub start: i64,
+    pub end: i64,
+    pub start_offset: i64,
+    pub end_offset: i64,
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct PlannedEdge {
```
Chris7, line 26: what's a planned edge?

gen-models/src/graph.rs
```diff
@@ -1,868 +0,0 @@
...
-
-pub struct ResolvedGraph {
-    pub graph: GenGraph,
-    pub interval_tree: IntervalTree<i64, NodeIntervalBlock>,
-    pub block_group_id: HashId,
-}
-
-/// Identify all edges leading to and from a provided node_id in a block_group and merge them into an existing GenGraph.
-/// Returns true if the graph was expanded, false if no new edges were added.
-pub fn expand(
```
Chris7, line 21: the new comments are AI written and in some places worse. It would be good to unify them again

Chris7: I mean if oyu look at the comments here vs. their new version, places like find_offset have been changed in ways we might not prefre

gen-models/graph-tests/tests/annotations.rs
```diff
@@ -0,0 +1,66 @@
+use gen_models::{
```
Chris7, line 1: is there a rationale for doing the dedicated test files? We don't have this style outside of integration tests.

gen-models/graph-tests/tests/migrated_model_graph.rs
```diff
@@ -0,0 +1,313 @@
+use gen_core::{GraphNode, HashId, NodeIntervalBlock, PathBlock, Strand};
```
Chris7, line 1: what's migrated_model_graph mean?

Chris7 (commented):
I want to take another look but submit initial comments now. 

The goal isn't clear to me. Is the goal for gen-models to not rely at all on gen-graph? The initial structure was gen-graph being an extension to petgraph -- have things that require opinions that petgraph can't do (such as how to merge graphs), models specific to our graph (Node/Edge definitions), and algorithms. This seems to complicate and expand gen-graph significantly (it was only like ~1k LoC) and make it a library that works on top of gen-models.

I know we've talked a bit about how dependency orders forced some structures we didn't love (essentially gen-diff). It would be good to have the reference frame of why this structure.

dkhofer:
🤦 Sorry for not writing a description.  I wrote up my reasons for doing this and a description of the changes, please take a look.  I'll work on addressing your other comments

Chris7:
Ok, so the core goal is to isolate database interactivity. The reason for things like region/etc. that did not use SQL being in gen-models was because they are Gen domain data. To me, gen-models is our domain mapping and the storage backend was an implementation detail. This makes gen-models explicitly defined by how it stores data instead of the domains the models are meant to represent.

I'm fine redefining the crate responsibilities -- ultimately the subcrates were made to improve compilation times. What are your thoughts on keeping gen-models as containing domain data, with persistence (serialization/deserialiation of the models) being a separate crate like gen-models-doltlite? We have traits.rs already defining what needs to be done for things like query generics. get_or_create/create/etc. can be standardized the same way with model specific queries being traits as well.

PR #241, opened 2026-08-09
Track completed clone and pull operations

Chris7:
This adds in a model tracking remote operations. The goal is tracking operations to ensure they are complete and resumable when they are not. If a user has an interrupted pull for example where an asset transfer is interrupted but the graph transfer has finished, it will resume with the asset transfer.

### DIFF
--- a/gen-models/migrations/config/01-initial/up.sql
+++ b/gen-models/migrations/config/01-initial/up.sql
@@ -12,6 +12,30 @@ CREATE TABLE remotes (
     url TEXT NOT NULL
 ) STRICT;
 
+CREATE TABLE remote_operations (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    remote_name TEXT NOT NULL,
+    branch_name TEXT NOT NULL,
+    operation TEXT NOT NULL CHECK(operation IN ('clone', 'pull')),
+    from_commit TEXT,
+    assets_transfer_checkpoint TEXT,
+    to_commit TEXT,
+    started_at TEXT NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    completed_at TEXT,
+    failed_at TEXT,
+    CHECK(completed_at IS NULL OR failed_at IS NULL),
+    CHECK(completed_at IS NULL OR (
+        to_commit IS NOT NULL
+        AND assets_transfer_checkpoint IS NOT NULL
+        AND assets_transfer_checkpoint = to_commit
+    )),
+    FOREIGN KEY(remote_name) REFERENCES remotes(name) ON DELETE CASCADE
+) STRICT;
+
+CREATE UNIQUE INDEX remote_operations_pending
+ON remote_operations(remote_name, branch_name)
+WHERE completed_at IS NULL AND failed_at IS NULL;
+
 CREATE TABLE remote_branch (
    remote_name TEXT,
    name TEXT,
--- a/gen-models/src/assets.rs
+++ b/gen-models/src/assets.rs
@@ -8,6 +8,7 @@ use std::{
 };
 
 use gen_core::{DoltHashId, HashId, Sha256Hash, Workspace, calculate_hash};
+use indexmap::IndexMap;
 use opendal::{blocking, services};
 use rusqlite::{
     Row, ToSql, named_params, params,
@@ -188,8 +189,12 @@ pub struct AssetRef {
 
 pub struct Assets;
 
-enum AssetView {
+/// Selects whether an asset-history query preserves every version or the final workspace view.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum AssetView {
+    /// Retains every local asset version in the selected history.
     Cumulative,
+    /// Retains only the asset selected for each logical path at the upper commit.
     Materialized,
 }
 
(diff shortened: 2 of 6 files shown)


gen-models/migrations/config/01-initial/up.sql
```diff
@@ -12,6 +12,30 @@ CREATE TABLE remotes (
     url TEXT NOT NULL
 ) STRICT;
 
+CREATE TABLE remote_operations (
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+    remote_name TEXT NOT NULL,
+    branch_name TEXT NOT NULL,
+    operation TEXT NOT NULL CHECK(operation IN ('clone', 'pull')),
```
Chris7, line 19: a follow up adds push

gen-models/src/assets.rs
```diff
@@ -307,23 +322,22 @@ impl AssetRef {
         Ok(())
     }
 
-    fn get_assets_at(
+    fn get_assets_by_commit(
```
Chris7: refactored this to return the commit at which an asset is introduced at. get_assets_at below uses this function to collapse this to just the assetref view.

gen-models/src/assets.rs
```diff
@@ -307,23 +312,36 @@ impl AssetRef {
         Ok(())
     }
 
-    fn get_assets_at(
+    /// Returns assets grouped by commit.
+    ///
+    /// Commits are in history order with commits without assets having an empty vector.
+    pub fn get_assets_by_commit(
```
Chris7, line 318: i renamed this function to get_assets_by_commit so it's an ordered hashmap by commit and all assets for that. This lets us get assets for a commit range in commit order. Here this is useful for being able to know the next batch of assets to download for resuming.

src/commands/remote/operations.rs
```diff
@@ -841,25 +857,49 @@ fn transfer_file_remote_assets(
     Ok(())
 }
 
+/// Get the assets between two commits and excludes the already checkpointed commit.
+fn get_remaining_assets_to_transfer(
```
Chris7, line 861: get_assets_by_commit is inclusive, this is just a helper function to kick out the first entry which we have already successfully transferred.

src/commands/remote/operations.rs
```diff
@@ -868,92 +908,144 @@ fn transfer_assets(
...
+                destination_logical_path,
+                &url,
+            )? {
+                warn_asset_conflict(workspace, &asset, destination_logical_path, &conflict_path)?;
+            }
+        }
+        complete_commit(&commit_hash)?;
     }
-    Ok(upload_receipts)
+    Ok(Vec::new())
```
Chris7, line 1042: receipts are only for pushes

gen-models/migrations/config/01-initial/up.sql
dkhofer: :+1:

gen-models/src/assets.rs
dkhofer: :+1:

gen-models/src/operations.rs
```diff
@@ -936,6 +1077,143 @@ mod tests {
...
+                RemoteOperationKind::Pull,
+                Some(&advanced_commit),
+            )
+            .expect("should resume pull operation");
+
+            assert_eq!(resumed, operation);
+            assert_eq!(resumed.from_commit, Some(original_commit));
+            assert_eq!(resumed.assets_transfer_checkpoint, Some(original_commit));
+
+            let mut resumed = resumed;
```
dkhofer: This could be declared mut from the start?

gen-models/src/operations.rs
```diff
@@ -936,6 +1077,143 @@ mod tests {
...
+            resumed
+                .set_destination(config, &advanced_commit)
+                .expect("should record pull destination");
+            resumed
+                .advance_assets_transfer_checkpoint(config, &advanced_commit)
+                .expect("should record pull asset checkpoint");
+            resumed
+                .complete(config)
+                .expect("should complete pull operation");
```
dkhofer, line 1183: It'd be good to verify the completed status in the database, by confirming completed_at is set, and that failed_at is null

gen-models/src/operations.rs
```diff
@@ -936,6 +1077,143 @@ mod tests {
...
+            )
+            .expect("should begin clone operation");
+            operation
+                .set_destination(config, &destination_commit)
+                .expect("should record graph destination");
+
+            operation
+                .complete(config)
+                .expect_err("should require assets to reach graph destination");
+        }
```
dkhofer, line 1259: Would be good to confirm completed_at isn't set, and I think failed_at would be set here?

src/commands/remote/operations.rs
dkhofer: :+1:

src/commands/remote/operations.rs
dkhofer: :+1:

src/commands/remote/operations.rs
```diff
@@ -1490,6 +1642,123 @@ mod tests {
         )
     }
 
+    fn serve_pull_api(
```
dkhofer, line 1645: Convenience methods like these have been in multiple recent PRs.  Would it make sense to make a general convenience function that takes a function defining the desired behavior of the server for specific tests, and returns the url + handle?

src/commands/remote/operations.rs
```diff
@@ -1948,6 +2246,310 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_execute_pull_transfers_assets_after_graph_db_is_synced() {
+        // Test that is we have transfered a graph.db, on a subsequent pull will we transfer assets. This mimics
```
dkhofer: This comment is kind of garbled, should be something like "Test that if we have transferred a graph.db, on a subsequent pull we will transfer assets"

dkhofer (approved):
LGTM, comments are non-blocking

src/commands/remote/operations.rs
Chris7: yeah, after all is done i want a followup to consolidate these and the little webservers

PR #242, opened 2026-08-10
Fix nondeterministic test failure, and a couple warnings

dkhofer:


src/views/annotations.rs
```diff
@@ -898,20 +898,30 @@ mod tests {
...
                         "HTTP/1.1 200 OK\r\nContent-Length: {}\r\nConnection: close\r\n\r\n",
                         contents.len()
                     )
-                    .expect("should write response headers");
-                    stream
-                        .write_all(&contents)
-                        .expect("should write response body");
+                    .and_then(|()| stream.write_all(&contents));
+                    if let Err(error) = response {
+                        assert!(
```
dkhofer, line 916: This fixes a nondeterministic test failure I was seeing locally.  I don't fully understand why a broken connection was causing it, but this seems to work

Chris7 (approved):
it seems like it hides the error instead by not using expect. Maybe we're just handling transient pipe disconnects that are happening? In anycase seems fine if it keeps the tests green.

PR #243, opened 2026-08-11
Add push tracking

Chris7:
Expands the previous operation tracking to support pushes. This includes a tracking id provided by the server that coordinates pushes across separate clients. The server (for now, maybe we can figure out concurrency better), issues a transfer id when a client requests a capability. The client can use that transfer id when pushing to effectively have a lock on the remote while its working.


dkhofer (approved):
LGTM, seems like a straightforward continuation of the pull/clone work

PR #244, opened 2026-08-12
Populate assets before materialization

Chris7:
This is a much smaller PR than it appears. Most of the code is new tests. It fixes our download workflow to what I thought it was -- download to .gen/assets and then materialize the file to the workspace if needed. Resumable downloads are attempted if the server supports them and downloads are restarted if they aren't or the asset is corrupted. I used the term `versioned` for assets in .gen/assets and materialized for what ends up outside the versioned path. I use this term because it reminds me of materialized views in sql.

### DIFF
--- a/src/commands/remote/operations.rs
+++ b/src/commands/remote/operations.rs
@@ -32,12 +32,12 @@
 //! GenHub. For an HTTP(S) repository remote, GenHub can return presigned URLs for the complete
 //! branch asset history. Gen uses the previous and destination commits to transfer only newly
 //! required versions. Push verifies each selected local file against its recorded checksum before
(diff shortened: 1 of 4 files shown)


src/commands/remote/operations.rs
```diff
@@ -776,13 +994,14 @@ mod tests {
         fs,
         io::{Cursor, Read as _, Write as _},
         net::TcpListener,
+        path::PathBuf,
```
Chris7, line 997: all the rest from here out are tests

dkhofer: :+1:

dkhofer (approved):
LGTM.  I'm fine with "materialized" describing "files that the user actually sees"

PR #245, opened 2026-08-12
Transfer versioned assets for file remotes

Chris7:
This adds in the same asset transfer stuff we just did for http:// remotes but ensures file:// remote urls are also supported. Note: this is a file:// REPOSITORY url, not an asset uri. So stuff like a NAS based server users sync with.

### DIFF
--- a/tests/history_cli_tests.rs
+++ b/tests/history_cli_tests.rs
@@ -20,7 +20,6 @@ use gen_models::{
     assets::AssetRef,
     block_group::BlockGroup,
     collection::Collection,
-    db::GraphConnection,
     history::{
         HistoryStore,
         dolt::{DoltHistoryStore, status_rows},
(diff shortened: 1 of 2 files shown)


tests/history_cli_tests.rs
```diff
@@ -2178,13 +2177,6 @@ mod remotes {
             .unwrap_or_else(|error| panic!("should resolve {reference}: {error}"))
     }
 
-    fn raw_graph_status(repo_root: &Path) -> Vec<gen_models::history::dolt::DoltStatusRow> {
```
Chris7, line 2181: we have a few things removed that were around schema mismatches -- this isn't a consideration here because we don't have the situation where branches have divergent schema.

dkhofer: :+1: Good

dkhofer (approved):
LGTM!

PR #247, opened 2026-08-15
Checkout materialize

Chris7:
We can currently push to a file:// remote, but there is no way to materialize files outside of .gen/assets. This adds in that missing piece.


src/commands/remote/operations.rs
```diff
@@ -675,10 +675,11 @@ fn copy_versioned_asset(
...
 ///
-/// HTTP downloads and `file://` copies call this only after the checksum-addressed version exists.
-/// Historical versions have no logical destination and remain solely in `.gen/assets`; selected
-/// versions replace a known prior version or are copied beside an unknown local file as a conflict.
-fn materialize_versioned_asset(
+/// HTTP downloads and `file://` copies pass the path produced by their storage phase. Checkout
+/// resolves that same checksum-addressed path before calling this function. Historical versions
+/// have no logical destination and remain solely in `.gen/assets`; selected versions replace a
+/// known prior version or are copied beside unknown local content as a conflict.
+pub(crate) fn materialize_versioned_asset(
```
Chris7, line 680: making these public so they can be used by checkout

tests/history_cli_tests.rs
```diff
@@ -2603,6 +2752,20 @@ mod remotes {
             resolved_ref_hash(local_repo_dir.path(), "main"),
             "post-push fetch should update the accepted tracking ref"
         );
+        let remote_fasta_path = remote_repo_dir.path().join(logical_path);
```
Chris7, line 2756: this adds a test on the remote side that when checking out post-push, the assets pushed are materialized

gen-models/src/assets.rs
```diff
@@ -393,7 +393,9 @@ impl AssetRef {
                                 historical_assets.created_on \
                                 {history_depth} \
                          FROM bounded_ancestry \
-                         JOIN dolt_history_gen_asset_refs AS historical_assets \
+                         JOIN dolt_history_gen_asset_refs( \
```
Chris7, line 396: this is actually the important part here. I'm treating the comment above as useful for LLMs b/c it wrote this query. But anyways -- dolt_history_gen_asset_refs needs a query parameter to scope the commits to return -- otherwise it will always use the main branch.

dkhofer: 👍

src/commands/remote/operations.rs
dkhofer: 👍

tests/history_cli_tests.rs
dkhofer: 👍

dkhofer (approved):
LGTM!

PR #249, opened 2026-08-17
Benchmark script for many small vcf updates

dkhofer:
I was trying to load some related virus genomes by repeatedly updating with vcfs containing the variants between each genome and the reference.  It went way slower than I expected, so I did some benchmarking and profiling.  This benchmarking script captures the basic workflow.

I'm seeing a couple things.  The initial one is that there's an incremental slowdown for each vcf update.  By the time it gets to the 100th update, it takes about a third of a second with the release build, even though the vcf files I've seen have had only a few dozen rows, with point mutations.  I had codex do some profiling and it looks like doltlite is the problem.  It apparently leaves the data as WAL entries inside the db, which grow on each update, making the timing proportional to O(n^2).  Codex made some changes to detect when the WAL entries got beyond a 64 MB threshold and compact them, which helps, but that feels a bit hacky.  I'll post those changes in a separate PR.

I went back to the last pre-doltlite commit and tried it against that, and it was much faster.  No incremental slowdowns, and my script was processing ~30/sec.  (Could speed that up more because half of the time was generating an alignment and a vcf for each genome.)

The other thing is, the disk usage for a database with 100 vcf updates is ~270 MB.  I feel 2.7 MB per update is way too much.  My ideal goal would be to load all ~9M genomes into a repo and share it.  I'd be fine if it took up say 1 KB or maybe more, but beyond that there's no way I'd want to share the repo currently, it'd be several terabytes.  I checked and the pre-doltlite gen's db used much less disk space, I think adding 300 KB per update or thereabouts.  Still not great but somewhat more reasonable.

It'd be good to improve the timings and disk space.


dkhofer:
The wal compaction change is here: https://github.com/genhub-bio/gen/pull/250

PR #250, opened 2026-08-17
Compact unindexed wal files if they get too big

dkhofer:
These are the changes that help with the slow timings I saw with https://github.com/genhub-bio/gen/pull/249

### DIFF
--- a/gen-models/src/migrations.rs
+++ b/gen-models/src/migrations.rs
@@ -21,6 +21,8 @@ enum MigrationError {
 }
 
 pub fn run_migrations(conn: &mut Connection) {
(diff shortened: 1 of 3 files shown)


gen-models/src/migrations.rs
```diff
@@ -21,21 +21,31 @@ enum MigrationError {
 }
 
 pub fn run_migrations(conn: &mut Connection) {
+    #[cfg(feature = "profiling")]
```
dkhofer, line 24: I had codex do some profiling, don't know if we want to keep these or not

Chris7:
guessing we can close this?

dkhofer:
Yep

PR #251, opened 2026-08-17
Fix build

dkhofer:
The rust compiler on my mac was getting deeper into recursion than it liked when doing type checking.  Changing the input parameter from an iterable to a list fixed it


Chris7 (approved):
I saw this change somewhere else on my end too. LGTM.

PR #252, opened 2026-08-18
Create a link for indices and other derived assets to their parent asset.

Chris7:
This is a generalization of assets with linked assets, such as indices. For example a .bai for a .bam and a .fai/.gzi for a bgzipped fasta file


dkhofer (approved):
LGTM

PR #253, opened 2026-08-18
Thread Workspace through sequence consumers

Chris7:
Cosmetic change where workspace is passed to all sequence creation. This is so on-disk operations can scope the workspace to find the correct file in a later PR


dkhofer (approved):
LGTM

PR #254, opened 2026-08-18
Back shallow sequences with AssetRefs

Chris7:
Currently we use a direct filepath for Sequence files to store their files on-disk instead of in the database. This changes it to leverage the AssetRef model to store a pointer to on-disk sequence files. This centralizes our file fetching and lets us leverage that for associated files such as indices without adding a bunch of extra columns to this model. It also ensures whatever security mesasures we have around AssetRefs are kept by Sequence so you can't pass `/etc/some_secret_stuff` as a filepath.

There's a bunch of kind of code churn resulting from splitting the fasta index work into a subsequent PR. I marked these parts so you can gloss over them.

Failing test resolves w/ the next Pr that fixes fasta indices

### DIFF
--- a/gen-models/src/annotations.rs
+++ b/gen-models/src/annotations.rs
@@ -20,10 +20,7 @@ use crate::{
     file_types::FileTypes,
     gen_models_capnp::{annotation, annotation_group, annotation_group_sample},
     history::{HistoryStore, dolt::DoltHistoryStore},
-    operations::{
-        FileAddition, OperationAssetRecord, OperationFile, OperationInfo, OperationSummary,
-        track_operation_assets,
-    },
+    operations::{FileAddition, OperationFile, OperationInfo, OperationSummary, track_asset_refs},
     traits::Query,
 };
 #[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
@@ -831,23 +828,28 @@ pub fn add_annotation_file(
     let summary = message
         .map(str::to_string)
         .unwrap_or_else(|| format!("Add annotation file {path}"));
-    let mut tracked_assets = vec![OperationAssetRecord {
-        file_addition: &file_addition,
-        role: AssetRole::Annotation,
-        logical_path: Some(&annotation_logical_path),
+    let created_on = chrono::Utc::now()
+        .timestamp_nanos_opt()
+        .expect("should create annotation asset timestamp");
+    let mut tracked_assets = vec![AssetRef::from_file_addition(
+        &file_addition,
+        AssetRole::Annotation,
+        Some(&annotation_logical_path),
         name,
-        upstream_asset_ref_id: None,
-    }];
+        None,
+        created_on,
+    )];
     if let Some((index_file_addition, index_logical_path, index_name)) = prepared_index.as_ref() {
-        tracked_assets.push(OperationAssetRecord {
-            file_addition: index_file_addition,
-            role: AssetRole::AnnotationIndex,
-            logical_path: Some(index_logical_path),
-            name: index_name.as_deref(),
-            upstream_asset_ref_id: Some(&annotation_asset_ref_id),
-        });
+        tracked_assets.push(AssetRef::from_file_addition(
+            index_file_addition,
+            AssetRole::AnnotationIndex,
+            Some(index_logical_path),
+            index_name.as_deref(),
+            Some(&annotation_asset_ref_id),
+            created_on,
+        ));
     }
-    track_operation_assets(
+    track_asset_refs(
         graph_conn,
         Some(&log_id),
         &OperationKind::AnnotationFile,
--- a/gen-models/src/assets.rs
+++ b/gen-models/src/assets.rs
@@ -276,6 +276,55 @@ impl Query for OperationAsset {
 }
 
 impl AssetRef {
+    /// Return the versioned store path for an asset.
+    ///
+    /// Logical paths describe the materialized workspace view and can point at a newer version.
+    /// Sequence readers use this path so they cannot follow repository metadata outside the
+    /// repository or silently switch to another asset version.
+    pub fn versioned_store_path(
+        &self,
+        workspace: &Workspace,
+    ) -> Result<PathBuf, FileAdditionError> {
+        let checksum = self.checksum.ok_or_else(|| {
+            FileAdditionError::ChecksumError(format!(
+                "asset {} has no checksum for repository lookup",
+                self.id
+            ))
+        })?;
+        let repo_root = LocalAssetUri::canonicalize_or_normalize(&workspace.repo_root()?);
+        let asset_path = workspace
+            .asset_dir()?
+            .join(<dyn AssetUri>::from_uri(&self.uri).hashed_filename(&checksum));
+        if !asset_path.exists() {
+            return Err(FileAdditionError::FileNotFound(
+                asset_path.display().to_string(),
+            ));
+        }
+        let resolved_path = LocalAssetUri::canonicalize_or_normalize(&asset_path);
+        if !resolved_path.starts_with(&repo_root) {
+            return Err(FileAdditionError::PathOutsideRepo {
+                path: resolved_path,
+                repo_root,
+            });
+        }
+        Ok(resolved_path)
+    }
+
+    /// Opens the asset identified by this reference.
+    ///
+    /// Local references resolve through the repository's content-addressed asset store. Remote
+    /// references retain their URI and are opened lazily through the configured storage backend.
+    pub fn reader(&self, workspace: &Workspace) -> Result<blocking::StdReader, FileAdditionError> {
+        if LocalAssetUri::is_local_path_or_file_uri(&self.uri) {
+            let asset_path = self.versioned_store_path(workspace)?;
+            return OpenDalLocation::from_absolute_path(&asset_path)
+                .map_err(opendal_file_addition_error)?
+                .reader();
+        }
+
+        <dyn AssetUri>::from_uri(&self.uri).reader(workspace)
+    }
+
     pub fn id_hash(
         uri: &str,
         file_type: &str,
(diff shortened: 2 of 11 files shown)


gen-models/src/operations.rs
```diff
@@ -93,6 +93,43 @@ impl OperationFile {
         self
     }
 
+    /// Retains an operation input and constructs the immutable asset reference that identifies it.
+    pub fn prepare_asset_ref(
```
Chris7, line 97: this consolidates stuff that happened for operation tracking below. The reason for it is sequences need an AssetRef, which was previously made during operation saving. This makes it so we can get the AssetRef earlier to associate with the sequence model. The operation saving code was refactored to use this as well to avoid having 2 code paths making assetrefs from operation files.

gen-models/src/annotations.rs
```diff
@@ -831,23 +828,28 @@ pub fn add_annotation_file(
     let summary = message
         .map(str::to_string)
         .unwrap_or_else(|| format!("Add annotation file {path}"));
-    let mut tracked_assets = vec![OperationAssetRecord {
-        file_addition: &file_addition,
-        role: AssetRole::Annotation,
-        logical_path: Some(&annotation_logical_path),
+    let created_on = chrono::Utc::now()
```
Chris7, line 831: there's a bunch of little reoganizations for asset saving that makes assetref creation happen outside of the operation saving, so places can reference assetrefs prior to the operation being saved.

gen-models/src/annotations.rs
```diff
@@ -831,23 +828,28 @@ pub fn add_annotation_file(
...
+            index_file_addition,
+            AssetRole::AnnotationIndex,
+            Some(index_logical_path),
+            index_name.as_deref(),
+            Some(&annotation_asset_ref_id),
+            created_on,
+        ));
     }
-    track_operation_assets(
+    track_asset_refs(
```
Chris7, line 852: track_operation_assets is gone, now its just track_asset_refs which persists that data. It's no longer just tied to an operation.

gen-models/src/assets.rs
```diff
@@ -301,6 +350,37 @@ impl AssetRef {
...
+                file_addition.checksum.as_ref(),
+                &role,
+                logical_path,
+                name,
+                upstream_asset_ref_id,
+            ),
+            uri: file_addition.asset_uri.clone(),
+            file_type: file_type.to_string(),
+            checksum: file_addition.checksum,
+            size: None,
```
Chris7, line 375: we never actually meaningfully use this field, might be useful to simply remove.

gen-models/src/operations.rs
```diff
@@ -153,119 +190,68 @@ pub fn commit_operation_summary(
...
         &operation_summary.summary,
-        &tracked_assets,
+        &prepared_assets,
     )?;
 
     let commit_result = history_store.commit_all(&operation_summary.summary);
     commit_result.map_err(OperationError::from)
 }
 
-pub(crate) struct OperationAssetRecord<'a> {
```
Chris7, line 206: no longer needed now that we don't tether assets to operations for creation

gen-models/src/sequence.rs
```diff
@@ -1,47 +1,79 @@
...
-use crate::{db::GraphConnection, gen_models_capnp::sequence, traits::*};
+use crate::{
+    assets::{AssetRef, AssetUri, LocalAssetUri},
+    db::GraphConnection,
+    gen_models_capnp::sequence,
+    traits::Query,
+};
 
-#[derive(Clone, Debug, Eq, Hash, PartialEq, Deserialize, Serialize)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
```
Chris7, line 19: eq/hash/partialeq are manually implemented now

gen-models/src/sequence.rs
```diff
@@ -1,47 +1,79 @@
...
     NoSequence(),
-    #[error("A filepath must have an accompanying sequence name.")]
-    FilepathMissingSequenceName(),
+    #[error("An asset reference must have an accompanying sequence name.")]
+    AssetMissingSequenceName(),
     #[error("Sequence length must be specified.")]
     MissingSequenceLength(),
     #[error("Sequence cache lock poisoned: {0}")]
     CachePoisoned(String),
-    #[error("Invalid indexed sequence region '{name}': {reason}")]
```
Chris7, line 43: gets readded back in a later PR

gen-models/src/sequence.rs
```diff
@@ -249,29 +294,20 @@ impl<'a> NewSequence<'a> {
...
-            external_sequence: !self.file_path.unwrap_or("").is_empty(),
+            external_sequence: self.asset_ref_id.is_some(),
+            asset_ref: None,
+            workspace: None,
+            asset_resolution_error: None,
         })
     }
 }
 
-#[cached(key = "String", convert = r#"{ format!("{}", path) }"#)]
```
Chris7, line 259: gets readded back in a later PR

gen-models/src/sequence.rs
```diff
@@ -249,29 +294,20 @@ impl<'a> NewSequence<'a> {
...
-}
-
-#[cached(key = "String", convert = r#"{ format!("{}", path) }"#)]
-fn fasta_gzi_index(path: &str) -> Option<gzi::Index> {
-    let index_path = format!("{path}.gzi");
-    if fs::metadata(&index_path).is_ok() {
-        return Some(gzi::fs::read(&index_path).unwrap());
-    }
-    None
+fn asset_extension(asset_ref: &AssetRef) -> Option<String> {
```
Chris7, line 307: we need this so we can get `.fa` from `.fa.gz`

gen-models/src/sequence.rs
```diff
@@ -361,78 +401,41 @@ pub fn cached_sequence(
         .map_err(|err| SequenceError::CachePoisoned(err.to_string()))?;
 
     let mut sequence: Option<String> = None;
-    if let Some(index) = fasta_index(file_path) {
```
Chris7, line 364: this all gets put back in for the next PR that handles indices for assets

gen-models/src/sequence.rs
```diff
@@ -449,6 +452,77 @@ impl Sequence {
             .any(|part| part.eq_ignore_ascii_case("circular"))
     }
 
+    pub fn query_by_ids<T>(
```
Chris7, line 455: reimplemented this trait to handle the assetref join to avoid N+1 queries

gen-models/src/sequence.rs
```diff
@@ -476,12 +569,29 @@ impl Sequence {
         stmt.execute(params![hash]).unwrap();
     }
 
-    pub fn query_by_blockgroup(conn: &GraphConnection, block_group_id: &HashId) -> Vec<Sequence> {
-        Sequence::query(
-            conn,
-            "select sequences.* from block_group_edges bge left join edges on bge.edge_id = edges.id left join nodes on (edges.source_node_id = nodes.id or edges.target_node_id = nodes.id) left join sequences on (nodes.sequence_hash = sequences.hash) where bge.block_group_id = ?1;",
-            params![block_group_id],
-        )
+    pub fn query_by_blockgroup(
```
Chris7, line 572: same thing about joining asset refs

src/imports/fasta.rs
```diff
@@ -41,16 +41,41 @@ pub fn import_fasta(
...
+    let mut checksum_handle = None;
+    let input: Box<dyn Read> = if shallow {
+        let created_on = i64::try_from(
+            SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .expect("should create sequence asset timestamp")
+                .as_nanos(),
+        )
+        .expect("should fit sequence asset timestamp in i64");
+        let sequence_asset_ref =
```
Chris7, line 55: this is basically why we needed the asset ref updates -- sequences are made here before operations so they need the asset ref id

gen-models/src/annotations.rs
dkhofer: 👍

gen-models/src/annotations.rs
dkhofer: 👍

gen-models/src/assets.rs
dkhofer: 👍

gen-models/src/operations.rs
dkhofer: 👍

gen-models/src/operations.rs
dkhofer: 👍

gen-models/src/sequence.rs
dkhofer: 👍

gen-models/src/sequence.rs
dkhofer: :+1:

gen-models/src/sequence.rs
dkhofer: :+1:

gen-models/src/sequence.rs
dkhofer: :+1:

gen-models/src/sequence.rs
dkhofer: :+1:

gen-models/src/sequence.rs
dkhofer: :+1:

gen-models/src/sequence.rs
dkhofer: :+1:

gen-models/src/sequence.rs
```diff
@@ -684,21 +820,33 @@ mod tests {
...
+            .save(context.graph().conn())
             .unwrap();
+        let seq = Sequence::query_by_ids(
+            context.graph().conn(),
+            context.workspace(),
+            &[sequence.hash],
+            None,
+        )
+        .remove(0);
+        fs::write(
```
dkhofer, line 847: What is this doing?  Is it showing that the sequence is in memory at this point?

src/imports/fasta.rs
dkhofer: :+1:

src/imports/fasta.rs
```diff
@@ -41,16 +41,41 @@ pub fn import_fasta(
...
+            operation_files[0].prepare_asset_ref(context.workspace(), created_on)?;
```
dkhofer: Unless I'm missing something, it's kind of weird to create a vec with one entry and then work with that one entry by referencing into the vec for it.  It'd be a little simpler to just create the operation file and then create a vec with it at the end, when it's needed

src/imports/fasta.rs
```diff
@@ -363,31 +470,109 @@ mod tests {
...
+            .into_iter()
+            .find(|sequence| sequence.asset_ref_id == Some(sequence_asset.id))
+            .expect("should resolve shallow sequence through remote AssetRef");
+        assert_eq!(
+            sequence
+                .get_sequence(2, 8)
+                .expect("should read remote sequence slice"),
+            "CGATCG"
         );
     }
```
dkhofer, line 574: This is maybe more of an integration test, but it might be good to have a test that imports a shallow fasta from outside of the repo, pushes the repo to a remote, clones it from the remote into a new directory, then confirms an error when trying to read the fasta.  That is the situation we're looking to reproduce, right?

dkhofer (approved):
A few questions and suggestions, but LGTM

gen-models/src/sequence.rs
Chris7: Its trying to show the local path doesn't matter but it fetches it from the versioned store

src/imports/fasta.rs
Chris7: its not, its testing a Remote as in the sequence comes from a Remote like ncbi/s3/etc.

src/imports/fasta.rs
Chris7: it was done b/c operation_files is appended to later. it can be a separate var and appended i guess at the end.

PR #256, opened 2026-08-21
Add native fetch command

Chris7:
We currently have no way to get remote branches locally. This adds the missing fetch functionality.


dkhofer (approved):
LGTM

PR #257, opened 2026-08-24
Store path edge IDs inline

Chris7:
Remove the path_edges table as store the edge_ids as an inline blob on path. This drastically improves path creation performance for GFAs where the reference path is broken up into many pieces.


gen-models/src/block_group.rs
```diff
@@ -222,7 +221,8 @@ impl<'a> PathCache<'a> {
             let conn = path_cache.conn;
             let new_path = Path::query(
                 conn,
-                "select * from paths where block_group_id = ?1 AND name = ?2",
```
Chris7, line 225: these are all updated to prevent querying for the large blob table when its not needed

dkhofer: :+1:

dkhofer (approved):
LGTM!

PR #258, opened 2026-08-24
Gfa edge index update

Chris7:
Updates the edge index to reduce disk usage. The primary key serves the same point as the index we have. Its replaced with 2 indices that are used by lookups.


dkhofer (approved):
LGTM

PR #259, opened 2026-08-24
Gfa misc improvements

Chris7:
Misc performance gains for importing GFAs

### DIFF
--- a/gen-models/src/block_group_edge.rs
+++ b/gen-models/src/block_group_edge.rs
@@ -2,7 +2,7 @@ use std::{collections::HashMap, rc::Rc};
 
 use gen_core::{HashId, calculate_hash, traits::Capnp};
 use indexmap::IndexSet;
-use rusqlite::{self, Row, ToSql, params, types::Value};
+use rusqlite::{self, Row, ToSql, types::Value};
 use serde::{Deserialize, Serialize};
 
 use crate::{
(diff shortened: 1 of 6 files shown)


gen-models/src/block_group_edge.rs
```diff
@@ -229,25 +229,31 @@ impl BlockGroupEdge {
         conn: &GraphConnection,
         block_group_id: &HashId,
         edge_ids: &[HashId],
+        history_ref: Option<&str>,
```
Chris7, line 232: adding history_ref to this function so it can get historical entries

gen-models/src/node.rs
```diff
@@ -70,6 +70,33 @@ pub enum NodeError {
 }
 
 impl Node {
+    /// Creates nodes in bounded batches, retaining any row that already has the same identifier.
```
Chris7, line 73: bulk node creation

gen-models/src/path.rs
```diff
@@ -143,8 +143,10 @@ impl Path {
             println!("Duplicate edge IDs detected in path creation");
         }
 
-        // All path edges must be in the path's block group
-        let augmented_edges = BlockGroupEdge::edges_for_block_group(conn, block_group_id, None);
+        // Validate against only the requested edges. A block group can be much larger than any
```
Chris7, line 281: restrict validation to edges that were added instead of all edges in the BG

src/imports/gfa.rs
```diff
@@ -92,25 +92,36 @@ pub fn import_gfa(
     let bar = progress_bar.add(get_time_elapsed_bar());
     bar.set_message("Parsing GFA");
     let gfa: Gfa<String, (), ()> = Gfa::parse_gfa_file(gfa_path.to_str().unwrap());
-    let mut sequences_by_segment_id: HashMap<&String, Sequence> = HashMap::new();
```
Chris7, line 94: we only use length, so store this instead of all the rest

src/imports/gfa.rs
```diff
@@ -246,7 +257,20 @@ pub fn import_gfa(
         edge_ids_by_data.insert(key, edge.id);
     }
 
-    let mut created_blockgroup_edges: IndexSet<HashId> = IndexSet::new();
+    // Every imported edge belongs to this block group regardless of how many paths traverse it.
```
Chris7, line 247: do all the BGEs here upfrnt instead of piecemeal

gen-models/src/block_group_edge.rs
dkhofer: :+1:

gen-models/src/node.rs
dkhofer: 👍

gen-models/src/path.rs
dkhofer: 👍

src/imports/gfa.rs
dkhofer: 👍

src/imports/gfa.rs
dkhofer: 👍

dkhofer (approved):
LGTM, happy to see the new bulk create methods

PR #255, opened 2026-08-18
Support indexed shallow FASTA assets

Chris7:
Utilized the previous asset index setup for fasta files. Resolves a handful of sequence churn from the last PR when index suppor wasn't quite there yet.

We need a follow up way to explicit add indices to asset refs for files where an index is added later.

PR #260, opened 2026-08-24
Fasta sequence indexes

Chris7:
Redo of #255 

Utilized the previous asset index setup for fasta files. Resolves a handful of sequence churn from the last PR when index suppor wasn't quite there yet.

We need a follow up way to explicit add indices to asset refs for files where an index is added later.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -44,6 +44,14 @@ being downloaded in full when the server supports range requests. Otherwise,
 the file is streamed into the cache and reused locally. Run `gen cache-clear` to
 remove the cache. Local repository files and `.gen/assets` are not affected.
 
+# FASTA import
+
+Use `gen import fasta <path-or-uri> --shallow` to keep sequence content in an
+asset rather than the graph database. Local assets are read from their immutable
+repository copy; remote assets remain remote. Supply known FASTA indexes with a
+repeatable `--index <path-or-uri>` option. For example, a BGZF FASTA can use both
+`--index reference.fa.gz.fai` and `--index reference.fa.gz.gzi`.
+
 # View diff
 
 Compare two commits or branches with:
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -46,9 +46,9 @@ remove the cache. Local repository files and `.gen/assets` are not affected.
 
 # FASTA import
 
-Use `gen import fasta <path-or-uri> --shallow` to keep sequence content in an
-asset rather than the graph database. Local assets are read from their immutable
-repository copy; remote assets remain remote. Supply known FASTA indexes with a
+Use `gen import fasta <path-or-uri> --shallow` to keep sequence data in an
+asset rather than the gen database. Local assets are read from their
+repository copy; remote assets remain remote. Supply known FASTA indices with a
 repeatable `--index <path-or-uri>` option. For example, a BGZF FASTA can use both
 `--index reference.fa.gz.fai` and `--index reference.fa.gz.gzi`.
 
--- a/gen-models/Cargo.toml
+++ b/gen-models/Cargo.toml
@@ -28,7 +28,7 @@ flate2 = "1.1.0"
 intervaltree = "0.2.7"
 itertools = "0.14.0"
 noodles = { version = "0.101.0", features = ["async", "bed", "bgzf", "core", "fasta", "gff", "gtf", "tabix", "vcf"] }
-opendal = { version = "0.56", features = ["blocking", "services-fs", "services-http", "services-s3", "services-gcs", "services-azblob"] }
+opendal = { version = "0.58", features = ["blocking", "services-fs", "services-http", "services-s3", "services-gcs", "services-azblob"] }
 petgraph = "0.6.5"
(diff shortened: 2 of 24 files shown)


gen-python/src/python_api/repository/imports.rs
```diff
@@ -36,18 +36,20 @@ impl PyRepository {
...
-                            PyRuntimeError::new_err(format!("Failed to import '{}': {e}", filename))
-                        }
-                    })?;
+                let operation_summary = import_fasta(
+                    ctx,
+                    &filename,
+                    &collection,
+                    &sample,
+                    shallow,
+                    &[],
```
Chris7, line 45: there's a bunch of these passing an empty array for index files

src/imports/fasta.rs
```diff
@@ -54,13 +55,36 @@ pub fn import_fasta(
...
         if let Some(checksum) = sequence_asset_ref.checksum {
-            operation_files[0] = operation_files[0].clone().set_checksum_override(checksum);
+            operation_file = operation_file.set_checksum_override(checksum);
         }
         AssetRef::create(conn, &sequence_asset_ref)
             .map_err(gen_models::errors::FileAdditionError::DatabaseError)?;
         sequence_asset_ref_id = Some(sequence_asset_ref.id);
+
+        let mut index_locations = indexes.to_vec();
+        for index_extension in ["fai", "gzi"] {
```
Chris7, line 67: autofind indices

gen-models/src/assets.rs
```diff
@@ -314,15 +318,20 @@ impl AssetRef {
     ///
     /// Local references resolve through the repository's content-addressed asset store. Remote
     /// references retain their URI and are opened lazily through the configured storage backend.
-    pub fn reader(&self, workspace: &Workspace) -> Result<blocking::StdReader, FileAdditionError> {
-        if LocalAssetUri::is_local_path_or_file_uri(&self.uri) {
+    pub fn reader(
+        &self,
+        workspace: &Workspace,
+    ) -> Result<BufReader<blocking::StdReader>, FileAdditionError> {
```
Chris7, line 324: switch to a buffered reader. This does look-ahead reads to minimize the number of IO calls and improves performance.

gen-models/Cargo.toml
```diff
@@ -28,7 +28,7 @@ flate2 = "1.1.0"
 intervaltree = "0.2.7"
 itertools = "0.14.0"
 noodles = { version = "0.101.0", features = ["async", "bed", "bgzf", "core", "fasta", "gff", "gtf", "tabix", "vcf"] }
-opendal = { version = "0.56", features = ["blocking", "services-fs", "services-http", "services-s3", "services-gcs", "services-azblob"] }
+opendal = { version = "0.58", features = ["blocking", "services-fs", "services-http", "services-s3", "services-gcs", "services-azblob"] }
```
Chris7, line 31: bumped this while testing some reader options for performance gains

docs/commands.md
```diff
@@ -44,6 +44,14 @@ being downloaded in full when the server supports range requests. Otherwise,
 the file is streamed into the cache and reused locally. Run `gen cache-clear` to
 remove the cache. Local repository files and `.gen/assets` are not affected.
 
+# FASTA import
+
+Use `gen import fasta <path-or-uri> --shallow` to keep sequence content in an
```
dkhofer: I might reword the firs sentence as "to keep sequence data in an asset file rather than the gen database".  Data instead of content, asset file instead of asset (do bio people know what an asset is?), and gen instead of graph to make it clearer we'd be copying and storing the data.  Or maybe "gen graph database".

gen-models/src/assets.rs
dkhofer: 👍 Cool

gen-models/src/sequence.rs
```diff
@@ -401,30 +433,91 @@ pub fn cached_sequence(
...
+                        .sequence()
+                        .as_ref(),
+                )
+                .map_err(|err| SequenceError::Utf8(err.to_string()))?
+                .to_string(),
             );
-            break;
+        }
+    }
+    if sequence.is_none() {
```
dkhofer, line 496: Could this just be an else clause continuing the if statement that checks if the asset has a fasta index?

dkhofer: If so, there might be a little more refactoring around setting the sequence in each block (the code looks almost the same in each case), but seems a bit tricky

gen-models/Cargo.toml
dkhofer: :+1:

gen-python/src/python_api/repository/imports.rs
dkhofer: :+1:

src/imports/fasta.rs
dkhofer: :+1:

dkhofer (approved):
LGTM, just minor feedback

gen-models/src/sequence.rs
Chris7: I kind of like how it is now being a failthrough in case there are weird setups, like someone with a .fai file + a bgz fasta w/o a gzi index will not read the sequence in that case but this check will get it.

PR #261, opened 2026-08-25
In memory path validation

Chris7:
Adds a path validation that uses the already created edges. On the e. coli GFA import this dropped import time from 418s to 120s.

### DIFF
--- a/gen-models/src/path.rs
+++ b/gen-models/src/path.rs
@@ -15,8 +15,14 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 use crate::{
-    block_group_edge::BlockGroupEdge, db::GraphConnection, edge::Edge, errors::QueryError,
-    gen_models_capnp::path as PathCapnp, node::Node, sequence::SequenceError, traits::*,
+    block_group_edge::BlockGroupEdge,
+    db::GraphConnection,
+    edge::{Edge, EdgeData},
+    errors::QueryError,
+    gen_models_capnp::path as PathCapnp,
+    node::Node,
+    sequence::SequenceError,
+    traits::*,
 };
 
 #[derive(Clone, Debug, Eq, Hash, PartialEq, Deserialize, Serialize)]
@@ -35,6 +41,8 @@ pub enum PathError {
     Missing(String),
     #[error("Duplicate entry with uuid: {0}")]
     Duplicate(String),
+    #[error("Invalid path: {0}")]
+    Invalid(String),
     #[error("Problem querying for path: {0}")]
     Query(#[from] QueryError),
     #[error("Problem loading sequence for path: {0}")]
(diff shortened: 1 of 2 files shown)


src/imports/gfa.rs
```diff
@@ -231,20 +230,8 @@ pub fn import_gfa(
 
     let gen_bar = progress_bar.add(get_time_elapsed_bar());
     gen_bar.set_message("Creating Gen Objects");
-    let edge_ids = Edge::bulk_create(conn, &edges.into_iter().collect::<Vec<EdgeData>>());
-
-    let saved_edges = Edge::query_by_ids(conn, &edge_ids, None);
```
Chris7, line 236: we already have this data, no need to query it back out

gen-models/src/path.rs
```diff
@@ -35,6 +41,8 @@ pub enum PathError {
     Missing(String),
     #[error("Duplicate entry with uuid: {0}")]
     Duplicate(String),
+    #[error("Invalid path: {0}")]
+    Invalid(String),
```
Chris7, line 45: changed from asserts to an error type

gen-models/src/path.rs
```diff
@@ -201,65 +209,108 @@ impl Path {
         Edge::query_by_ids(conn, &edge_ids, history_ref)
     }
 
+    fn validate_ordered_edges(edge_data: &[EdgeData]) -> Result<(), PathError> {
```
Chris7, line 212: I decomposed the different parts of validation to have a shared in-memory path. The major difference is one path fetches data from the database whereas the other path uses the in-memory data it has already fetched.

gen-models/src/path.rs
```diff
@@ -201,65 +209,108 @@ impl Path {
...
-                edge1.id,
-                edge2.id,
-            );
-        }
+        let ordered_edge_data = edge_ids
+            .iter()
+            .map(|edge_id| {
+                edges_by_id.get(edge_id).cloned().ok_or_else(|| {
+                    PathError::Invalid(format!(
+                        "edge {edge_id} is not in the block group ({block_group_id})"
```
Chris7, line 304: this should never hit given the above check but its simply condition handling

gen-models/src/path.rs
dkhofer: Oh, great!

gen-models/src/path.rs
dkhofer: 👍

gen-models/src/path.rs
dkhofer: 👍

src/imports/gfa.rs
dkhofer: 👍

dkhofer (approved):
LGTM!

PR #262, opened 2026-08-27
Use versioned assets for annotation loading

Chris7:
Bias annotation loading use the asset store instead of materialized paths. This prevents loading the wrong annotations for a commit in case a file has been uploaded/replaced and lets places like servers which do not materialize files to use this loader.

### DIFF
--- a/src/views/annotations.rs
+++ b/src/views/annotations.rs
@@ -1565,11 +1565,9 @@ mod tests {
         assert!(!cached_bytes.is_empty());
     }
 
-    #[test]
-    fn test_annotations_use_versioned_assets() {
-        // Add annotations and then delete the local file, ensuring assets come from the versioned asset store
+    fn setup_versioned_annotation_assets()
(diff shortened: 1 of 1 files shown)


src/views/annotations.rs
```diff
@@ -1542,4 +1564,131 @@ mod tests {
...
+            &annotation_path,
+            Some("gff3"),
+            Some(&index_path),
+            Some("genes"),
+            Some("add genes"),
+            AnnotationFileChecksumOverrides::default(),
+        )
+        .expect("should commit annotation file");
+        fs::remove_file(&annotation_path).expect("should remove original annotation source");
+        fs::remove_file(&index_path).expect("should remove original annotation index");
```
dkhofer: Maybe also have a test where the files are altered but not deleted?

dkhofer (approved):
LGTM, just a minor suggestion

PR #265, opened 2026-08-31
Bulk_create for annotations

Chris7:
Adds bulk_create for annotations. I noticed the e. coli genbank file was taking a while to import. This drops import time from a few mins to 2.45s on my tests.


gen-models/src/annotations.rs
```diff
@@ -67,16 +67,13 @@ impl AnnotationGroup {
...
-            Err(rusqlite::Error::SqliteFailure(err, _details))
-                if err.code == rusqlite::ErrorCode::ConstraintViolation =>
-            {
-                AnnotationGroup::get_by_id(conn, &name.to_string(), None)
-                    .ok_or_else(|| rusqlite::Error::QueryReturnedNoRows.into())
-            }
-            Err(err) => Err(err.into()),
-        }
+        let mut statement = conn.prepare(
+            "INSERT INTO annotation_groups (name) VALUES (?1) ON CONFLICT(name) DO NOTHING;",
```
Chris7, line 71: This explicitly handles conflicts instead of using IGNORE, keeping stuff like CHECK and other safeguards in place

dkhofer: :+1:

dkhofer: Are there other places where we should be doing this instead of using IGNORE (that are currently using IGNORE)?

dkhofer (approved):
Awesome, LGTM

gen-models/src/annotations.rs
Chris7: yeah, i think we should make all the places w/ blanket ignores do it and maybe refactor all our get_or_creates

PR #268, opened 2026-09-03
Sample diff command

Chris7:
I split out cosmetic from actual changes in this PR. https://github.com/genhub-bio/gen/pull/268/changes/bbb37f07b9e618950f57815334b53323ec767aa3 has the changes, feel free to comment there

This generalizes the graph diff builder a bit more to support diffs of a blockgroup between 2 samples.

The language of the prior version was confusing, so I also tried to make it more specific. A big one is removing the stage/staged terms, which made it appear as if there were multiple graphs or something similar. It was meant to be like git staged but it simply means the input so that was updated.

For the actual change, I used the term `base` and `query`. If you have better ones, let me know. `source` and `target` became very confusing as we have source/target nodes/edges. Base is meant to be baseline (but less letters) and query is what you want to search with. Reference and sample were similarly rejected as you can mix and match comparing against a reference or a sample.

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -54,6 +54,28 @@ repeatable `--index <path-or-uri>` option. For example, a BGZF FASTA can use bot
 
 # View diff
 
+Compare one sequence graph between two samples with:
+
+```sh
+gen view <graph> --query <query> --base <base>
+```
+
+The base establishes the reference for the comparison. For example,
+`gen view m123 --query foo --base unknown` shows how `foo` differs from
+`unknown`. A sequence present only in the query is shown as an addition in
+green, while a sequence present only in the base is shown as a removal in red.
+Add `--full` to open the comparison in the full-screen viewer.
+
+To export a sample comparison as GFA, use the same endpoint names:
+
+```sh
+gen diff --query <query> --base <base> --gfa <output.gfa>
+```
+
+The older `--sample1` and `--sample2` names remain aliases for compatibility.
+
+## Revision diff
+
 Compare two commits or branches with:
 
 ```sh
--- a/gen-diff/src/graph.rs
+++ b/gen-diff/src/graph.rs
@@ -37,7 +37,7 @@
 //! A parent is selected only to make an added or removed child graph
 //! biologically meaningful. The changed rows would show a child's entire graph
 //! being removed or created. Thus, comparing to the parent shows the edits that
-//! introduced in the child. `load_block_group_stage_graphs` therefore
+//! introduced in the child. `load_block_group_comparison_graphs` therefore
 //! keeps two concepts separate:
 //!
 //! - `source_block_group` and `target_block_group` preserve actual membership
@@ -63,7 +63,7 @@
 //!
 //! # Graph-building approach
 //!
-//! Graph construction has three stages:
+//! Graph construction has three steps:
 //!
 //! 1. Load the selected endpoint edge sets and convert edge coordinates into
 //!    [`GraphNode`] sequence slices. Coordinates are slices of the backing
@@ -122,7 +122,7 @@ pub struct DiffChange {
 impl DiffChange {
     /// Creates the neutral annotation used for graph elements shared by both endpoints.
     ///
-    /// Stage builders and merge helpers use this constructor so unchanged
+    /// Input builders and merge helpers use this constructor so unchanged
     /// context has neither a change kind nor misleading operation attribution.
     pub const fn unchanged() -> Self {
         Self {
@@ -133,7 +133,7 @@ impl DiffChange {
 
     /// Creates an annotation from the change kind and attribution collected from Dolt.
     ///
-    /// Stage-graph construction uses this common constructor before the source
+    /// Input-graph construction uses this common constructor before the source
     /// and target annotations are consolidated into the rendered diff graph.
     pub const fn new(kind: DiffChangeKind, operation: Option<DoltHashId>) -> Self {
         Self { kind, operation }
@@ -265,7 +265,7 @@ pub(crate) struct BlockGroupDiffInputs<'a> {
 /// Builds the renderable diff for one block group identified by table-diff discovery.
 ///
 /// `build_dolt_operation_diff` calls this once per affected block group. It
-/// resolves normalized endpoint stage graphs, unifies them, and omits a
+/// resolves normalized endpoint input graphs, unifies them, and omits a
 /// source-and-target block group when the resulting graph has no visible
 /// changes, avoiding empty entries in command and TUI output.
 pub(crate) fn build_block_group_diff(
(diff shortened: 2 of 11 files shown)


gen-diff/src/graph.rs
```diff
@@ -37,7 +38,7 @@
 //! A parent is selected only to make an added or removed child graph
 //! biologically meaningful. The changed rows would show a child's entire graph
 //! being removed or created. Thus, comparing to the parent shows the edits that
-//! introduced in the child. `load_block_group_stage_graphs` therefore
```
Chris7, line 40: so i hated the stage nomenclature. To me it sounded like there were multiple stages. Stage just meant an intermediate state (like "staged" in git). So these terms were cleaned up/removed.

gen-diff/src/graph.rs
```diff
@@ -265,46 +266,51 @@ pub(crate) struct BlockGroupDiffInputs<'a> {
 /// Builds the renderable diff for one block group identified by table-diff discovery.
 ///
 /// `build_dolt_operation_diff` calls this once per affected block group. It
-/// resolves normalized endpoint stage graphs, unifies them, and omits a
+/// resolves normalized endpoint input graphs, unifies them, and omits a
```
Chris7, line 318: this is what i mean, a stage graph sounds like its some special type of graph vs. just the input graph

gen-diff/src/graph.rs
```diff
@@ -265,46 +266,51 @@ pub(crate) struct BlockGroupDiffInputs<'a> {
...
 /// source-and-target block group when the resulting graph has no visible
 /// changes, avoiding empty entries in command and TUI output.
 pub(crate) fn build_block_group_diff(
     graph_conn: &GraphConnection,
     block_group_id: HashId,
     inputs: &BlockGroupDiffInputs<'_>,
 ) -> Result<Option<BlockGroupDiff>, OperationDiffError> {
-    let Some(stage_graphs) = load_block_group_stage_graphs(graph_conn, block_group_id, inputs)?
+    let Some(comparison_graphs) =
```
Chris7: another rename, i'm calling stuff comparisons. Hoping that is more future proof and clear

gen-diff/src/graph.rs
```diff
@@ -265,46 +266,51 @@ pub(crate) struct BlockGroupDiffInputs<'a> {
...
+        load_block_group_comparison_graphs(graph_conn, block_group_id, inputs)?
     else {
         return Ok(None);
     };
-    let graph = build_unified_diff_graph(&stage_graphs, inputs.default_operation);
```
Chris7, line 280: the source and target graphs were made explicit arguments for build_unified_diff_graph. Nothing material

docs/commands.md
```diff
@@ -54,6 +54,28 @@ repeatable `--index <path-or-uri>` option. For example, a BGZF FASTA can use bot
...
+The base establishes the reference for the comparison. For example,
+`gen view m123 --query foo --base unknown` shows how `foo` differs from
+`unknown`. A sequence present only in the query is shown as an addition in
+green, while a sequence present only in the base is shown as a removal in red.
+Add `--full` to open the comparison in the full-screen viewer.
+
+To export a sample comparison as GFA, use the same endpoint names:
+
+```sh
+gen diff --query <query> --base <base> --gfa <output.gfa>
```
Chris7, line 72: This was carried in. If you want to keep sample1/sample2, lmk. base/query seems better to which one belongs on what side of the diff.

gen-diff/src/graph.rs
```diff
@@ -378,45 +409,41 @@ fn load_block_group_stage_graphs(
...
-        &stage_operation_change_lookup.source_node_operation_by_id,
-        &stage_operation_change_lookup.source_edge_operation_by_id,
-    );
-    let reconstructed_target_graph = build_stage_graph_from_nodes(
-        target_nodes,
-        &target_spans,
-        &target_edges,
-        &stage_operation_change_lookup.target_node_operation_by_id,
-        &stage_operation_change_lookup.target_edge_operation_by_id,
+    let graph = build_diff_graph_from_edges(
```
Chris7, line 438: since nodes come from edges, we can actually get rid of the nodes call above and simplify this code.

gen-diff/src/graph.rs
```diff
@@ -379,43 +426,37 @@ fn load_block_group_comparison_graphs(
         target_effective_block_group_id,
         target_effective_history_ref,
     );
-    let source_spans = spans_from_edges(&source_edges);
-    let target_spans = spans_from_edges(&target_edges);
+    let source_input = EdgeDiffInput::unattributed(&source_edges);
```
Chris7, line 429: spans get calculated on the EdgeDiffInput constructor now. They were always used, so this effectively caches it.

gen-diff/src/graph.rs
dkhofer: :+1:

docs/commands.md
dkhofer: I'm good with `base`, but instead of `query` I'd be more comfortable with `newer` or `changed` or something.  In my mind `query` means something related to text search, which I don't think fits well with this usage

gen-diff/src/graph.rs
dkhofer: :+1:

gen-diff/src/graph.rs
dkhofer: :+1:

gen-diff/src/graph.rs
dkhofer: :+1:

gen-diff/src/graph.rs
dkhofer: :+1:

gen-diff/src/graph.rs
```diff
@@ -711,42 +783,39 @@ fn span_node_ids(spans: &HashSet<GraphNode>) -> HashSet<HashId> {
...
-        .source_graph
-        .edge_changes
-        .keys()
-        .chain(stage_graphs.reconstructed_target_graph.edge_changes.keys())
-        .copied()
+    let mut merged_edge_keys = source_graph
+        .edges()
+        .map(|(edge_key, _)| edge_key)
+        .chain(target_graph.edges().map(|(edge_key, _)| edge_key))
         .collect::<Vec<_>>();
```
dkhofer: Why not collect this into a HashSet, since it looks like they need to be unique and don't need to be sorted (which the next two lines do)?

gen-diff/src/graph.rs
```diff
@@ -711,42 +783,39 @@ fn span_node_ids(spans: &HashSet<GraphNode>) -> HashSet<HashId> {
...
-        .node_changes
-        .keys()
-        .chain(stage_graphs.reconstructed_target_graph.node_changes.keys())
-        .copied()
+    let mut merged_nodes = source_graph
+        .nodes()
+        .chain(target_graph.nodes())
         .collect::<Vec<_>>();
     merged_nodes.sort();
     merged_nodes.dedup();
```
dkhofer: Actually, seems like these could be collected into a HashSet too?

src/diffs/gfa.rs
```diff
@@ -33,36 +33,35 @@ pub fn gfa_sample_diff(
     workspace: &Workspace,
     collection_name: &str,
     filename: &PathBuf,
-    from_sample_name: &str,
-    to_sample_name: &str,
+    query_name: &str,
```
dkhofer: This flips how the names were meant to be applied.  I intended "from" to mean "base" and "to" to mean "query".  Similarly, "source" should be "base" and "target" should be query

dkhofer (approved):
Overall LGTM, just some quibbles about naming and other minor stuff

gen-diff/src/graph.rs
Chris7: i agree, changing

src/diffs/gfa.rs
Chris7: Thanks for catching this. I thought tests would catch this, but they only look at the collection of resulting sequences (the gfa), not what should be added/removed. So for a gfa export because it has no indicators both directions are the same output wise

PR #269, opened 2026-09-07
Bump rusqdoltlite to 0.40.18 to fix HTTPS remote transfers

bobvh:
libdoltlite-sys 0.38.16 caused handshakes to an https:// GenHub remote to fail with "could not connect to remote", this Cargo.lock change bumps to the version where it has been fixed.


dkhofer (approved):
LGTM, but if the tests keep breaking please take a look

PR #274, opened 2026-09-08
Support starting a new branch at a given operation

bobvh:
The basic idea is to feel more like git:

|#|Before                                                                                               |After                                                                                   |Git equivalent                                            |
|-|-----------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------|----------------------------------------------------------|
|1|`gen checkout -b <name> <hash>` → creates `<name>` at current HEAD, silently ignoring `<hash>`       |`gen checkout -b <name> <hash>` → creates `<name>` starting at `<hash>`                 |`git checkout -b <new-branch> <start-point>`              |
|2|`gen checkout -b <existing-name> <hash>` → silently checks out the existing branch, ignoring `<hash>`|`gen checkout -b <existing-name> <hash>` → errors: branch already exists                |`git checkout -b <existing-branch> ...` → errors similarly|
|3|*(not possible — no way to pass a start point)*                                                      |`gen branch --create <name> <start-point>` → creates `<name>` starting at that operation|`git branch <name> <start-point>`                         |
|4|*(n/a)*                                                                                              |`gen branch <start-point>` without `--create` → errors                                  |*(no git equivalent)*                                     |
|—|`gen checkout <hash>` (no `-b`) → detached HEAD error                                                |unchanged                                                                               |git allows this; gen intentionally doesn't                |

### DIFF
--- a/tests/history_cli_tests.rs
+++ b/tests/history_cli_tests.rs
@@ -1499,6 +1499,105 @@ mod revision_views {
     }
 
     #[test]
+    fn test_checkout_branch_create_from_hash_starts_new_branch_at_operation() {
+        let repo_dir = tempdir().expect("should create temp repo directory");
+        let (_, feature_update_hash, feature_sample_name) =
+            setup_repo_with_feature_only_update(repo_dir.path());
+
+        assert_success(
+            &run_gen(
+                repo_dir.path(),
+                &["checkout", "--branch", "from-feature", &feature_update_hash],
+            ),
+            "checkout -b from a feature commit hash should succeed",
+        );
+
+        let samples = run_gen(repo_dir.path(), &["list-samples"]);
+        assert_success(&samples, "list-samples on from-feature should succeed");
+        let stdout = String::from_utf8_lossy(&samples.stdout);
+        assert!(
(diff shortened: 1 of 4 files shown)


bobvh:
We could do a more thorough change to the commands too and make --create implicit, or have "gen branch" list branches like git does.

tests/history_cli_tests.rs
```diff
@@ -1498,6 +1498,105 @@ mod revision_views {
...
+        );
+        let stderr = String::from_utf8_lossy(&checkout.stderr);
+        assert!(
+            stderr.contains("already exists"),
+            "checkout -b should explain that the branch already exists rather than silently reusing HEAD: {stderr}"
+        );
+    }
+
+    #[test]
+    fn test_branch_create_with_start_starts_new_branch_at_operation() {
```
Chris7, line 1547: worth trying a hash that isn't at a branch tip

Chris7 (approved):
lgtm. One test suggestion.

tests/history_cli_tests.rs
bobvh: Done

PR #277, opened 2026-09-09
Py branches

bobvh:
Adding remote commands to the python bindings (see below) and made this change that affects the cli (I ran into it when testing out the new methods).

Before, you couldn't clone to an existing path, even if it's an empty directory. This gets awkward when you want to clone a repo in a temp dir created specifically to hold the data. Now you can specify an existing path as long as it's empty. The other change that was needed then was to prevent that path from getting removed entirely on failure. (copying files to the path during cloning that then fails will result in those files being lost, I think that's acceptable for now?)

On the python bindings themselves:
- removes the user-facing transaction context manager entirely (per Chris' request earlier)
- adds Branch, Operation and Asset objects
- adds clone, remote config, push, pull, fetch
- there's a new notebook to demonstrate all of this


bobvh:
The windows test failing is from trying to cleanup a repo while the python Repository still exists. Testing a fix now.

src/commands/clone.rs
bobvh, line 1: Chris already looked at this. Conclusion was to mirror what git does.

PR #278, opened 2026-09-10
Add paginated lineage queries

Chris7:
For lazy loading of lineages

### DIFF
--- a/gen-models/src/lineage.rs
+++ b/gen-models/src/lineage.rs
@@ -282,6 +282,85 @@ pub trait SqlLineage: Sized {
         LineagePage { ids, has_more }
     }
 
+    /// Return ancestors ordered by minimum depth and ID with a global page bound.
+    ///
+    /// The page limits rows returned to the caller. The recursive traversal still visits the
+    /// bounded-depth subgraph so convergent paths can be deduplicated before the offset is
+    /// applied.
+    fn get_ancestors_page(
+        conn: &Connection,
+        child_id: &Self::Id,
+        max_depth: Option<usize>,
+        limit: usize,
+        offset: u32,
+        history_ref: Option<&str>,
+    ) -> LineagePage<Self::Id> {
+        let max_depth = max_depth.map(|depth| depth as i64);
+        let lineage_table_name =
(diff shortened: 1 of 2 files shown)


gen-models/src/lineage.rs
```diff
@@ -74,6 +83,196 @@ pub trait SqlLineage: Sized {
...
+            sql_table_name_with_history_ref(Self::TABLE_NAME, Some(Self::TABLE_NAME), history_ref);
+        let parent_table_name = sql_table_name_with_history_ref(
+            Self::PARENT_TABLE_NAME,
+            Some(Self::PARENT_TABLE_NAME),
+            history_ref,
+        );
+        let query = format!(
+            "SELECT parent.{parent_id_column}
+             FROM {parent_table_name} parent
+             WHERE NOT EXISTS (
```
Chris7, line 112: so -- why don't we lookup parent_id is null. If we have a join table, such as:
```
Entities

   id
  ━━━━━
   1
  ─────
   2
  ─────
   3

  Lineage

   parent_id    child_id
  ━━━━━━━━━━━  ━━━━━━━━━━
   1            2
```

looking up parent_id is null would not find any roots. This query supports both configurations

Chris7: A better way to think of it is looking at the table of potential parents, then excising those w/ children

gen-models/src/lineage.rs
```diff
@@ -158,9 +357,10 @@ pub trait SqlLineage: Sized {
                 SELECT
                     lineage.{child_column},
                     1,
-                    printf('|%s|', hex(lineage.{child_column}))
+                    printf('|%s|%s|', hex(:parent_id), hex(lineage.{child_column}))
```
Chris7, line 448: prevent cycles

gen-models/src/lineage.rs
dkhofer: :+1:

gen-models/src/lineage.rs
dkhofer: :+1:

dkhofer (approved):
LGTM

PR #284, opened 2026-09-15
Allow path edges to meet at the same coordinate

bobvh:
Path::validate_edges() rejected a path where consecutive edges land on
the same coordinate, which is a legitimate zero-width junction (e.g. a
route that enters and exits a deleted node's boundaries without
consuming any of its bases). Relax the check from "coordinates must
strictly increase" to "coordinates must not go backwards," and add a
test covering a path built through such a junction.


gen-models/src/path.rs
```diff
@@ -263,9 +263,11 @@ impl Path {
                     second_edge.source_node_id
                 )));
             }
-            if first_edge.target_coordinate >= second_edge.source_coordinate {
```
Chris7, line 266: should this actually be !=? This is really just saying is the connection point valid and the edge sorting valid.

Chris7 (commented):
should it be !=?

PR #290, opened 2026-09-17
Add conflict resolution for known cases

Chris7:
This resolves known conflicts. If we create the same edges/etc. in different branches, the created_on date  will differ while the rest of the columns are the same. This adds application level handling of known cases where conflicts will arise and handling them.

To do this we:
* Added a transaction handler (the conflicts are only resolvable in transactions)(
* Added handling for known cases where a resolution is known

General conflict resolution outside these cases are an error and a TBD


gen-models/src/history/dolt.rs
```diff
@@ -437,21 +437,146 @@ pub fn connect_branch(conn: &GraphConnection, branch_name: &str) -> SqlResult<()
...
-/// Merges a branch or commit reference into the active branch and returns the resulting hash.
-pub fn merge(conn: &GraphConnection, reference: &str) -> SqlResult<DoltHashId> {
-    let result = conn.query_row("SELECT dolt_merge(?1)", [reference], |row| {
-        row.get::<_, String>(0)
+fn resolve_created_on_conflicts(
+    conn: &GraphConnection,
+    conflicts: &[HistoryConflict],
+) -> SqlResult<()> {
+    // This handles known conflict cases automatically. The cases are:
+    // Created_on dates between otherwise identical can differ. Automatically use the older created_on date for a conflict
```
dkhofer, line 445: *otherwise identical objects

gen-models/src/history/dolt.rs
```diff
@@ -437,21 +437,146 @@ pub fn connect_branch(conn: &GraphConnection, branch_name: &str) -> SqlResult<()
...
+/// The `dolt_merge('--no-commit', reference)` outcomes are:
+///
+/// | Result | Meaning |
+/// | --- | --- |
+/// | `Text("Already up to date")` | No-op; return the current HEAD. |
+/// | `Text(commit hash)` | Fast-forward; return the resulting hash. |
+/// | `Integer(0)` | Clean non-fast-forward working set; requires `dolt_commit`. |
+/// | `SQLITE_ERROR` with data conflicts | Conflict rows materialized for selective resolution before commit. |
+///
+/// Callers must invoke this operation inside [`GraphConnection::with_transaction`] for the
```
dkhofer, line 534: Is there an easy way to mechanically enforce this?  Like setting a variable with GraphConnection that indicates if a transaction is open and then checking it here

gen-models/src/history/dolt.rs
```diff
@@ -437,21 +437,146 @@ pub fn connect_branch(conn: &GraphConnection, branch_name: &str) -> SqlResult<()
...
+            Ok(commit_hash) => return Ok(commit_hash),
+            Err(error) => {
+                return Err(rusqlite::Error::FromSqlConversionFailure(
+                    0,
+                    Type::Text,
+                    Box::new(error),
+                ));
+            }
+        },
+        Ok(Value::Integer(0)) => None,
```
dkhofer, line 560: I'm seeing two paths here that don't return something, this one that results in None and then a general error below.  Which one of these is expected to be the result if there are conflicts?  Maybe add a comment indicating it?

gen-models/src/history/dolt.rs
```diff
@@ -1368,18 +1499,325 @@ mod tests {
...
+            "an aborted merge must not advance the target branch"
+        );
+        assert_eq!(
+            conn.query_row(
+                "SELECT created_on FROM block_group_edges WHERE id = ?1",
+                [&base_block_group_edge_id],
+                |row| row.get::<_, i64>(0),
+            )
+            .expect("should read target block group edge after failed merge"),
+            3,
```
dkhofer, line 1727: This test made me realize that the comment for the conflict resolution method, `Automatically use the older created_on date for a conflict`, is incorrect, instead it takes the target created_on

dkhofer: So I think the comment needs to be updated

gen-models/src/db.rs
```diff
@@ -36,6 +36,60 @@ impl Deref for GraphConnection {
...
+    ///
+    /// An existing transaction remains owned by its caller. Dolt version-control statements can
+    /// seal a transaction opened by this helper themselves, so commit and rollback are
+    /// conditional on the connection still being in a transaction after the operation returns.
+    pub fn with_transaction<T>(
+        &self,
+        operation: impl FnOnce() -> rusqlite::Result<T>,
+    ) -> rusqlite::Result<T> {
+        let in_transaction = self.start_transaction()?;
+        let result = operation();
```
dkhofer, line 88: This is very ruby-block-like, the fact that functions can be passed around enables it

dkhofer: There other places that could use the syntax update to call this instead of starting/ending transactions, right?

gen-models/src/db.rs
```diff
@@ -297,4 +352,176 @@ mod tests {
...
+
+    #[test]
+    fn test_with_transaction_rolls_back_failed_operation() {
+        let conn = test_graph_connection(None).expect("should create graph database");
+        conn.execute("CREATE TABLE transaction_rows (value TEXT NOT NULL)", [])
+            .expect("should create transaction fixture table");
+
+        let error: rusqlite::Result<()> = conn.with_transaction(|| {
+            conn.execute(
+                "INSERT INTO transaction_rows (value) VALUES ('rolled back')",
```
dkhofer, line 388: I'm feeling kind of stupid but I'm not understanding why this statement is expected to generate an error

dkhofer (approved):
LGTM, some take-it-or-leave-it suggestions, plus some requests for more or improved comments

gen-models/src/history/dolt.rs
Chris7: There's a table above outlining the return values.

gen-models/src/history/dolt.rs
Chris7: yeah, we can check is_autocommit which means using sqlite's per statement transaction mode (the default)

gen-models/src/history/dolt.rs
Chris7: good catch, mentally i was thinking the target is "older" but not factually correct

gen-models/src/db.rs
Chris7: yeah, i think outer level wrappers will benefit from this

gen-models/src/db.rs
Chris7: it's good sql, its verifying that if we do have an error (which we immediately introduce to the function return) that we rollback

PR #293, opened 2026-09-17
Add blockgroup region search UI

Chris7:


gen-models/src/region.rs
```diff
@@ -499,6 +568,58 @@ impl ResolvedGenRegion {
         }
     }
 
+    pub fn graph_locus(
```
Chris7, line 590: converts the region to a GraphLocus, which is a Vec<GraphNodeSlices>, which is then used for knowing the actual portions of the graph covered by the region

gen-models/src/accession.rs
```diff
@@ -249,6 +249,25 @@ pub struct NewAccession {
 }
 
 impl Accession {
+    pub(crate) fn resolve_candidates(
```
Chris7, line 252: current resolver only returns 1 match. This makes it so places can choose a lookup that returns all entries. this is used by the search interface.

gen-models/src/region.rs
```diff
@@ -119,6 +120,72 @@ pub fn resolve(
     resolve_accession(region, conn, collection_name, sample_name)
 }
 
+pub fn resolve_all(
```
Chris7, line 123: complementary method to resolve that returns all matches instead of a precedence order

dkhofer: :+1:

gen-models/src/accession.rs
dkhofer: :+1:

gen-models/src/region.rs
dkhofer: :+1:

gen-models/src/region.rs
```diff
@@ -741,9 +884,81 @@ mod tests {
...
+                    node_id: HashId::convert_str("test-t-node"),
+                    range: Range { start: 2, end: 8 },
+                    strand: Strand::Reverse,
+                }],
+            },
+        )
+        .unwrap();
+        let region = ResolvedGenRegion::from_accession(&conn, &accession, 2, 5).unwrap();
+
+        // absolute:          0 1 2 3 4 5 6 7 8 9 10
```
dkhofer, line 907: Thanks for this comment

dkhofer (approved):
LGTM.  Is there an example repo you were using to test this out?  Could you post it or files + operations to build the repo?  I'd like to try it out

PR #297, opened 2026-09-20
Add all-sequences FASTA export support

bobvh:
I know we talked about exposing get_all_sequences before. For library design I do think it's necessary because there's no way to get your full library out in a fasta export. This PR adds an option all_sequences to the fasta export CLI command, python bindings. For the python api I also wanted an iterator so we don't blow up memory with massive strings, hence not using get_all_sequences directly.

Minimal changes to core code, but it's in some important files. It walks the graph but is futureproof wrt cycles (see test).


Chris7 (commented):
is there a reason get_all_sequences isn't being used here? It returns a lazy iterator iirc

bobvh:
@Chris7 : it returns a hashset of strings for all_simple_paths (see comment below)

gen-models/src/block_group.rs
bobvh, line 707: existing method (eager)

Chris7: Yes, you can template the return type though so the caller can decide the type and use a trait like extend. Or define our own trait w/ a common append method for vec/string

PR #299, opened 2026-09-21
Prevent pruning from disconnecting the graph when inserting next to a prior insert.

bobvh:
TL;DR: Inserting a node next to a previously inserted node (homozygous change not retaining the parent edge)
broke because of a zero-length anchor that would get pruned out.

Full:
Anchored edits that do not preserve the old edge add edit-site marker edges at their anchor positions. When the anchor is a node's first or last coordinate, the marker creates a zero-width junction instead of splitting any sequence. Once that marker is pruned, a later non-preserving edit chained onto the same boundary can end up disconnected from the path.

Markers at node boundaries are no longer added. Markers needed for internal splits and for live preserved routes are kept.

The regression test in `gen-models/src/region.rs` applies two non-preserving insertions, anchors the second at the end of the first, prunes the graph, and checks that both the second insertion and the path end are still reachable.


Chris7 (commented):
can you add an ascii diagram? i don't really get the situtation we're mapping here. Why is this in region as well?

bobvh:
It's in region.rs because the edits lean on the Accession machinery (ResolvedGen_region::plan_edges) , not the path coordinates. 

Ascii diagrams:

0) initial: 
`AAAAAAAAA`
1) 1st insertion: 
```
AAAA - TTTT - AAAA
   |__________|
```
3 new edges: A-T, T-A and the chrom index -2 marker edge right after A3 to split the A node

2) 2nd insertion right at the junction: 
```
AAAA - TTTT - CCCC - AAAA 
           |_|
```
3 new edges again, but since it's at the end of the T node, the marker edge right after T3 doesn't split the node but results in a zero-width junction after T.

Then your insert connects to that junction, which admittedly is weird but how I saw the edges were implemented.
```
AAAA - TTTT - [J] -  CCCC - AAAA 
                 |__________|
````


Maybe a more robust fix would be to prevent that junction to C connection, but i suspect that has to do the fact that all of those edges include T3.

bobvh:
select_junction_or_sequence_blocks is one of your recent additions, right? I think there may be something funky there that bleeds into this.

Chris7:
The junction node makes it so you can connect to the extremes of something without making a ton of new edges between the adjacent edges and then handling that iteratively

Chris7:
See #232

bobvh:
Replied over there since the screenshots are there too: https://github.com/genhub-bio/gen/pull/232#issuecomment-5784667443

bobvh:
tl;dr: need to weigh style points vs information content.

